PR #4, opened 2024-01-10
Return path to downloaded file in download_url

jashlearn:
This PR is largely a test of the gnssanalysis contribution processes but it also adds a nice feature of returning a path to the downloaded file in a download function.


ronaldmaj (approved):
Awesome, this is a nice addition, we can then use that path in functions upstream

ronaldmaj (approved):
Reasonable addition of a type hint to this PR. Approved :+1:

PR #7, opened 2024-01-15
NPI-3060 Remove duplicated epochs in SP3 reader (with warning)

treefern:
Duplicated epochs should **not** be present in SP3 files. A warning is now logged if any are found. Detecting and *removing* duplicates also allows comparisons to be run on less compliant data.

The actual duplicate check is done on the Pandas DataFrame with `Index.duplicated()`. A quick test suggests this executes in less than 1ms on a typical SP3 file, so is not a performance concern.

### DIFF
--- a/gnssanalysis/gn_io/sp3.py
+++ b/gnssanalysis/gn_io/sp3.py
@@ -149,11 +149,11 @@ def read_sp3(sp3_path, pOnly=True):
     sp3_df.attrs["path"] = sp3_path
 
     # Check for duplicate epochs, dedupe and log warning
-    duplicated_indexes = sp3_df.index.duplicated()
+    duplicated_indexes = sp3_df.index.duplicated() # Typically sub ms time. Marks all but first instance as duped.
     if duplicated_indexes.sum() > 0: # We have dupes
         first_dupe = sp3_df.index.get_level_values(0)[duplicated_indexes][0]
         logging.warning(
-            f"Duplicate epoch(s) found in SP3! "
+            f"Duplicate epoch(s) found in SP3 ({duplicated_indexes.sum()} additional entries, potentially non-unique). "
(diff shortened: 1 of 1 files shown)


gnssanalysis/gn_io/sp3.py
```diff
@@ -141,6 +147,18 @@ def read_sp3(sp3_path, pOnly=True):
 
     sp3_df.attrs["HEADER"] = parsed_header  # writing header data to dataframe attributes
     sp3_df.attrs["path"] = sp3_path
+
+    # Check for duplicate epochs, dedupe and log warning
+    duplicated_indexes = sp3_df.index.duplicated()
+    if duplicated_indexes.sum() > 0: # We have dupes
+        first_dupe = sp3_df.index.get_level_values(0)[duplicated_indexes][0]
```
ronaldmaj, line 154: Not sure if it is worth it to list out all indexes or rows that are copied out (to show the scale of the problem?)
Perhaps doing something like `sp3_df.index.get_level_values(0)[duplicated_indexes].sum() / 2` should reveal the number of entries that are copied? Not entirely sure on this so it would need to be tested

ronaldmaj (approved):
We have had a chat about this already, I think this is good to go in. There is one comment I've put in, but I think we can change things around in future if we feel we need knowledge on the number of duplicate entries

gnssanalysis/gn_io/sp3.py
treefern: Updated to output the count in the warning. By the doc, it's by default the number of entries which duplicate something already there. So a (possibly repeated) list of duplicated values. I've worded the warning to try and capture that.

ronaldmaj (approved):
Awesome, thanks for putting the count in - good to go :+1:

PR #8, opened 2024-01-15
Updated sp3 dupe check \w has_duplicates property

bmatv:
As `has_duplicates` is an already a property inside the sp3 dataframe it is literally free. You may argue that `index.duplicated()` before `if` is not a waste  as you still need it to compute `first dupe`, but I could speculate that 99% of sp3 files come without dupes.


treefern (approved):
Nice improvement :) This is also a bit more readable.

bmatv:
@treefern Originally I added a comment on `has_duplicates` to you branch but never clicked on `Finish your review` button so it was a pending [read as invisible] comment. But this branch is also a great test on how-to commit/merge etc when one is a poor contributor, not a maintainer :)

PR #9, opened 2024-01-18
NPI-3072 additional comments explaining why we read Sinex with read_fwf() and that field names do not map

treefern:


ronaldmaj (approved):
No functional code changes but really helpful comments so I think this can go in as is :+1: 
Not sure if you want to wait on @jashlearn or @bmatv to provide additional commentary around this, but looks good to me

bmatv:
@treefern @ronaldmaj  I love it! We might dump the slow `read_fwf` in the future for a faster parsing method but that method is to be discovered

treefern:
I assume `read_fwf` would be a bit faster (and more reliable) if we explicitly set column boundaries, but I haven't tested this.

I think we could also improve robustness and maybe speed, by adding logic to first trim to the data structure needed. Eg `+SOLUTION/APRIORI` through `-SOLUTION/APRIORI`

bmatv:
@treefern you could definitely measure this but from experience it won't help. One could make it marginally faster by not asking for STD column if it's not present but there is no way of telling from a header of the file or the block if STD is present or not... SINEX is a pain to read efficiently in python

PR #11, opened 2024-01-24
Move download functions from ginan/scripts/auto_download_PPP to gnssanalysis

aclel:
Download functions in auto_download_PPP have wider use. Proposing to incorporate into gnssanalysis.

### DIFF
--- a/gnssanalysis/filenames.py
+++ b/gnssanalysis/filenames.py
@@ -306,9 +306,12 @@ def nominal_span_string(span_seconds: float) -> str:
             unit = "W"
             span_unit_counts = int(span_seconds // gn_const.SEC_IN_WEEK)
     elif span_seconds >= gn_const.SEC_IN_WEEK:
-        if (span_seconds % gn_const.SEC_IN_WEEK) < gn_const.SEC_IN_DAY:
+        num_weeks = int(span_seconds // gn_const.SEC_IN_WEEK)
+        # IGS uses 07D to represent a week
+        # TODO: Handle JPL - uses 01W for a week
+        if (span_seconds % gn_const.SEC_IN_WEEK) < gn_const.SEC_IN_DAY and num_weeks > 1:
             unit = "W"
-            span_unit_counts = int(span_seconds // gn_const.SEC_IN_WEEK)
+            span_unit_counts = num_weeks
         else:
             unit = "D"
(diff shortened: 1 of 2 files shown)


gnssanalysis/filenames.py
```diff
@@ -306,9 +306,12 @@ def nominal_span_string(span_seconds: float) -> str:
             unit = "W"
             span_unit_counts = int(span_seconds // gn_const.SEC_IN_WEEK)
     elif span_seconds >= gn_const.SEC_IN_WEEK:
-        if (span_seconds % gn_const.SEC_IN_WEEK) < gn_const.SEC_IN_DAY:
+        num_weeks = int(span_seconds // gn_const.SEC_IN_WEEK)
```
ronaldmaj, line 309: This is a good idea, makes things a lot clearer for the `if` statement that comes next

gnssanalysis/gn_download.py
```diff
@@ -339,6 +345,7 @@ def get_install_crx2rnx(override=False, verbose=False):
             logging.info(f"crx2rnx already present in {_sys.path[0]}")
 
 
+# TODO: Deprecate in favour of the contextmanager?
```
ronaldmaj, line 348: We'll need to check where this is still used and replace it, but yep, I think we can definitely get rid of this after that

gnssanalysis/gn_download.py
```diff
@@ -445,6 +479,59 @@ def download_most_recent(
...
+    :param files: List of str filenames
+    :ftp_folder: Folder where the file is stored on the remote
+    :output_folder: Folder to store the output files
+    """
+    with _concurrent.futures.ThreadPoolExecutor() as executor:
+        # Wrap this in a list to force iteration of results and so get the first exception if any were raised
+        list(executor.map(download_file_from_cddis, files, _repeat(ftp_folder), _repeat(output_folder)))
+
+
+# TODO: Deprecate? Only supports legacy filenames
```
ronaldmaj, line 534: Yeah, I think once we have the other functions that allow `download_file_from_cddis` to carry out the expected file downloads, we can definitely get rid of this.

Other functions here I mean stuff like filename generation

ronaldmaj (approved):
I think this can go in as is - the functions to deprecate can also be removed once we do a bit of an audit of the rest of the codebase to make sure they are not used elsewhere

PR #12, opened 2024-02-12
NPI-3064 Fix SP3 output formatting of nodata values, to avoid writing NaNs to SP3 file

treefern:


gnssanalysis/gn_io/sp3.py
```diff
@@ -317,44 +340,51 @@ def clk_log(x):
     def prn_formatter(x):
         return f"P{x}"
 
+    # TODO NOTE
```
ronaldmaj, line 343: No note left here?

gnssanalysis/gn_io/sp3.py
```diff
@@ -365,7 +395,38 @@ def clk_std_formatter(x):
...
         out_buf.write("\n")
+
         # Format this epoch's values in the SP3 format and write to buffer
+
+        # First, we fill NaN and infinity values with the standardised nodata value for each column.
+        # NOTE: DataFrame.to_string() as called below, takes formatter functions per column. It does not, however
+        # invoke them on NaN values!! As such, trying to handle NaNs in the formatter is a fool's errand.
+        # Instead, we do it here, and get the formatters to recognise and skip over the already processed nodata values
+        
+        # POS nodata formatting
```
ronaldmaj, line 406: Awesome, this whole section reads well - pretty clear what is being done and in a logical manner 👍🏻

ronaldmaj (approved):
This is great! Thanks for putting this together so quickly, and I think you've structured really well - clear what is being done and it's done in a way that allows edits in the future. 
If this has been tested and works as expected, happy for this to go in 👍🏻

gnssanalysis/gn_io/sp3.py
treefern: Ah, I added this line just to more clearly draw attention to the large block of commentary already there.

Specifically the fact we're not precicely following the SP3 standard, as we have whitespace between columns. If we have reason to rework our SP3 writing in future, this feels like an important motivator to include in the discussion.

gnssanalysis/gn_io/sp3.py
```diff
@@ -35,10 +35,20 @@
 # File descriptor and clock
 _RE_SP3_HEAD_FDESCR = _re.compile(rb"\%c[ ]+(\w{1})[ ]+cc[ ](\w{3})")
 
+# Nodata ie NaN constants for SP3 format
+SP3_CLOCK_NODATA_STRING  = " 999999.999999" # Not used for reading, as fractional components are optional
```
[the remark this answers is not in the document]
treefern: I was also motivated to do it because I think it makes it clearer what is being done when a value is e.g. used several layers into some calculation logic.

PR #15, opened 2024-03-28
fixing stats + residual estimation

seballgeyer:
+ Fixing the helmert sign for orbit rotation. 
+ changes of stats for STD


ronaldmaj (approved):
Installed this locally and tested `orbq` on a couple SP3 files. Seems to work well with the -h option for both ECF and ECI. Can't say too much on the accuracy of the result (don't have something else to compare against), but happy for this to go in

PR #16, opened 2024-03-28
Update python-publish.yml

ronaldmaj:
Could not release due to Node.js deprecation issue. This site suggests the following should work: @v3 --> @v5
https://stackoverflow.com/questions/77897660/node-js-16-actions-are-deprecated-please-update-the-following-actions-to-use-no


seballgeyer (approved):
Makes sense, 
Only test is triggering  the pipeline.

PR #17, opened 2024-03-28
Update python-publish.yml

ronaldmaj:
Previous PR had both set at v5, reading into the previous article again, I might need keep the actions/checkout at v4 instead 
https://stackoverflow.com/questions/77897660/node-js-16-actions-are-deprecated-please-update-the-following-actions-to-use-no

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -20,7 +20,7 @@ jobs:
     name: Build and publish Python distributions to PyPI
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v5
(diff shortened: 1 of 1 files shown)


seballgeyer (approved):
Okay for testing.

ronaldmaj:
The example on the `actions/setup-python` repo seems to show exactly this in the example in the README:
https://github.com/actions/setup-python?tab=readme-ov-file#basic-usage

```
steps:
- uses: actions/checkout@v4
- uses: actions/setup-python@v5
  with:
    python-version: '3.10' 
- run: python my_script.py
```

PR #19, opened 2024-04-28
Correct SISRE calculation

TYTY666:
Dear gnssanalysis developers, hello! Thank you for your work.

We noticed that there seems to be some errors in the function used to calculate SISRE in the `gn_diffaux.py` file of gnssanalysis. We believe that when calculating the SISRE indicator, `clk_diff` should be multiplied by the speed of light and mutually offset with the radial orbit deviation.

We have corrected this issue. If you think our correction is correct, please accept our PR. If there are any questions, please give us feedback.

### DIFF
--- a/gnssanalysis/gn_diffaux.py
+++ b/gnssanalysis/gn_diffaux.py
@@ -500,7 +500,7 @@ def sisre(
     alpha, beta = _np.vsplit(coeffs_df.values, indices_or_sections=2)
 
     sisre = _pd.DataFrame(
-        data=_np.sqrt((alpha * radial + clk_diff) ** 2 + (along**2 + cross**2) / beta),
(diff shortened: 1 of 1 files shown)


ronaldmaj:
Hi [TYTY666](https://github.com/TYTY666),

Thanks so much for your PR. It's a really good catch and something we missed when creating these functions. 

I have put through a parallel PR in order to move the multiplication by the speed of light up to where the `clk_diff` variable is defined. This aligns better with how we currently calculate the clk values in the `diffclk` function

The one area we will still take some time to investigate is the minus sign you've introduced in `alpha * radial - clk_diff`. This could very well be correct but we need to double check this and verify that this is the way it should be.

We will comment here again on this PR if we do find something.

Thanks again for bringing our attention to this :)

EugeneDu-GA (approved):

seballgeyer (changes requested):
Waiting for other fixes before merging this one.

PR #20, opened 2024-05-06
NPI-3290 Clearer error on empty Sinex header

treefern:
Raise a clearer error (referencing the source file if possible) when an empty Sinex header is encountered.


jashlearn:
This looks pretty decent. Out of interest do you have an example of a failing file that we can look at? I'm just wondering about the error message. It somehow feels unlikely that we see a header line that contains `%=SNX` only (which is a true empty header line) compared to a file that is either completely empty or is missing the header and doesn't have `%=SNX` anywhere. In which case we might want to be doing a test immediately after the search for the start of the header and providing a "missing header" error message at that point.

ronaldmaj (approved):
A nice catch for the SNX header - happy for this to go in

PR #22, opened 2024-05-15
Fix errors in RINEX reading function

jashlearn:
The existing rinex reading functions unfortunately atrophied over time and there was an failure in the line that extracted the expected column count from the signal header information (maybe due to changes in pandas versions) and then also failed using datetime2j2000 due to the runtime type checking in that function.
I reworked the column detection and then removed the type checking in the datetime2j2000 function (that function still might need love and updated type hints but I started by just deleting the branching because it seemed to have the same code on both branches).
This go the function working on the data I originally cared about but testing on further random data revealed it failed for /LGOW00AUS_S_20212590000_01D_30S_MO.rnx obtained from GA's gnss-data and more generally wouldn't work for any RINEX file with more than two lines of signals in the header.
I entirely reworked the way the header reading code works and I think simplified it overall. I also did a little tidy-up in the code generally and it seems to work for a large random sampling of rinex (3) files pulled from gnss-data.
This code still doesn't handle rinex 2, it is a bit brittle in places, and a little funky and maybe slow in implementation (it does a lot of weird fiddling with strings to do the parsing process and I think it can probably all be replaced with block-by-block read_fwf with some post-processing) but it's better than it started.

### DIFF
--- a/gnssanalysis/gn_datetime.py
+++ b/gnssanalysis/gn_datetime.py
@@ -195,12 +195,6 @@ def datetime2gpsweeksec(array: _np.ndarray, as_decimal=False) -> _Union[tuple, _
 
 def datetime2j2000(datetime: _np.ndarray) -> _np.ndarray:
     """datetime64 conversion to int seconds after J2000 (2000-01-01 12:00:00)"""
-    if not isinstance(datetime, (_np.ndarray, _np.datetime64)):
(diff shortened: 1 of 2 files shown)


ronaldmaj (approved):
Looks good. I've tested this locally and after the new commits you've put through, works well in reading in RINEX files and outputting them as a dataframe.
Thanks for picking up on this and updating it for changes that have occurred in the dependent modules

PR #24, opened 2024-05-16
Sat exlusion

seballgeyer:
Adding option to exclude satellite in orbq. (same as in clkq)


ronaldmaj (approved):
Tested this on SP3 files, comparing our Ginan produced Rapid product against IGS rapid and IGS ultra-rapid. The reject capability works well, effectively the same as how it works in clocks. Also tried with Helmert transform applied - still worked fine.
Happy for this to go in

PR #25, opened 2024-05-21
Convert datetimes explicitly to int64 not int

jashlearn:
On at least one platform (Windows I think) I've seen while working with someone else  `int` can be a 32-bit integer and the conversion from numpy datetime64/timedelta64 objects fail to convert to 32-bit integers (fair enough due to overflow possibilities). But we don't care about the size of the integer, we just want the conversion to work, and so I've changed to explicitly converting to 64-bit integers.
There's generally also some suspicious stuff going on in gn_datetime but I didn't bother looking at it at this stage, just making the minimal changes needed at this stage.

### DIFF
--- a/gnssanalysis/gn_datetime.py
+++ b/gnssanalysis/gn_datetime.py
@@ -161,8 +161,8 @@ def datetime2yydoysec(datetime: _Union[_np.ndarray, _pd.Series]) -> _np.ndarray:
         )  # .astype("datetime64[Y]") called on ndarray will return 4-digit year, not YYYY-MM-DD as in case of Series
     datetime_Y = datetime.astype("datetime64[Y]")
     datetime_D = datetime.astype("datetime64[D]")
-    doy = _pd.Series((datetime_D - datetime_Y).astype(int).astype(str))
-    seconds = _pd.Series((datetime - datetime_D).astype("timedelta64[s]").astype(int).astype(str))
+    doy = _pd.Series((datetime_D - datetime_Y).astype("int64").astype(str))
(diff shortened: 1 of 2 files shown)


ronaldmaj (approved):
Had a quick test of this, seems to work fine on a couple `datetime` conversions I did with relevant functions:
```py
In [15]: dt = datetime(year=2024, month=5, day=28,hour=4,minute=37)
...
In [21]: datetime2j2000(np.array(dt))
Out[21]: 770143020

In [22]: datetime2gpsweeksec(np.array(dt))
Out[22]: (2316, 189420)
```
There are other errors that seem to be causing problems, but these aren't related to your change, e.g.:
```py
In [20]: datetime2yydoysec(np.array(dt))
UFuncTypeError                            Traceback (most recent call last)
Cell In[20], line 1
----> 1 datetime2yydoysec(np.array(dt))

File /data/review-repos/gnssanalysis/gnssanalysis/gn_datetime.py:165, in datetime2yydoysec(datetime)
    163 datetime_D = datetime.astype("datetime64[D]")
    164 doy = _pd.Series((datetime_D - datetime_Y).astype("int64").astype(str))
--> 165 seconds = _pd.Series((datetime - datetime_D).astype("timedelta64[s]").astype("int64").astype(str))
    166 yydoysec = (
    167     _pd.Series(datetime_Y.astype(str)).str.slice(2).values
    168     + ":"
   (...)
    171     + seconds.str.zfill(5).values
    172 )
    173 return yydoysec

UFuncTypeError: ufunc 'subtract' cannot use operands with types dtype('O') and dtype('<M8[D]')
```

Happy for this PR to go in

PR #26, opened 2024-05-30
NPI-3331 - Update to gnssanalysis for new packages

ronaldmaj:
We have been testing Python 3.12 and updating various python packages, including Pandas.

There were a number of errors that popped up and these needed to be sorted in this module before we could run our usual operational jobs. 

Therefore, this PR brings changes to various scripts to ensure proper functioning after update to pandas + other packages

The most important changes are:
- removing additional `+` from the `rec_str_list` variable in `meta2sting()`
- changing the `nd.fill()` function to fill with NATs rather than NANs in the `get_bad_sv_from_nanu_df()` function
- splitting out the transformation of `std_df` into individual components as it seems like the `pd.concat` operation isn't playing nicely with the dataframe `attrs`.

### DIFF
--- a/gnssanalysis/gn_io/sp3.py
+++ b/gnssanalysis/gn_io/sp3.py
@@ -327,6 +327,11 @@ def gen_sp3_content(sp3_df: _pd.DataFrame, sort_outputs: bool = False, buf: Unio
             ).astype(int)
 
         std_df = sp3_df["STD"]
+        #  Remove attribute data from this shallow copy of the Dataframe.
+        #  This works around an apparent bug in Pandas, due to the fact calling == on two Series produces a list
+        #  of element-wise comparisons, rather than a single boolean value. This list seems to break Pandas
+        #  concat() when it is invoked within transform() and tries to check if attributes match between columns
+        #  being concatenated.
         std_df.attrs = {}
         std_df = std_df.transform({"X": pos_log, "Y": pos_log, "Z": pos_log, "CLK": clk_log})
         std_df = std_df.rename(columns=lambda x: "STD_" + x)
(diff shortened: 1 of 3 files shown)


gnssanalysis/gn_io/sp3.py
```diff
@@ -316,25 +314,22 @@ def gen_sp3_content(
...
-            .transform({"X": pos_log, "Y": pos_log, "Z": pos_log, "CLK": clk_log})
-            .rename(columns=lambda x: "STD" + x)
-        )
+                _np.nan_to_num(_np.rint(_np.log(x) / _np.log(clk_base)), nan=SP3_CLOCK_STD_NODATA), 999  # Cap at 999
+            ).astype(int)
+
+        std_df = sp3_df["STD"]
+        std_df.attrs = {}
+        std_df = std_df.transform({"X": pos_log, "Y": pos_log, "Z": pos_log, "CLK": clk_log})
+        std_df = std_df.rename(columns=lambda x: "STD" + x)
```
treefern, line 332: This would be slightly cleaner as:
```suggestion
        std_df = std_df.rename(columns=lambda x: "STD_" + x)
```

gnssanalysis/gn_io/sp3.py
```diff
@@ -316,25 +314,22 @@ def gen_sp3_content(
...
-        std_df = (
-            sp3_df["STD"]
-            .transform({"X": pos_log, "Y": pos_log, "Z": pos_log, "CLK": clk_log})
-            .rename(columns=lambda x: "STD" + x)
-        )
+                _np.nan_to_num(_np.rint(_np.log(x) / _np.log(clk_base)), nan=SP3_CLOCK_STD_NODATA), 999  # Cap at 999
+            ).astype(int)
+
+        std_df = sp3_df["STD"]
+        std_df.attrs = {}
```
treefern, line 330: Adding some explanation of our rationale here, for future explorers / maintainers. I may not have described the situation entirely accurately, feel free to amend.

```suggestion
        #  Remove attribute data from this shallow copy of the Dataframe.
        #  This works around an apparent bug in Pandas, due to the fact calling == on two Series produces a list
        #  of element-wise comparisons, rather than a single boolean value. This list seems to break Pandas
        #  concat() when it is invoked within transform() and tries to check if attributes match between columns
        #  being concatenated.
        std_df.attrs = {}

treefern (approved):

treefern (approved):

PR #27, opened 2024-06-03
Npi 3290 clearer error on empty sinex header extras

treefern:


gnssanalysis/gn_io/sinex.py
```diff
@@ -25,17 +25,21 @@
...
 def _get_snx_header(path_or_bytes):
+    # Helper for error logging
+    def file_desc(pb) -> str:
+        return "Source file: unknown (passed as bytes)" if isinstance(path_or_bytes, bytes) else "Source file: "+str(path_or_bytes)
+    
     # read line that starts with %=SNX and parse it into dict
     snx_bytes = _gn_io.common.path2bytes(path_or_bytes)
     header_begin = snx_bytes.find(b"%=SNX")
+    if (header_begin == -1):
+        raise ValueError("Sinex header missing!", file_desc(path_or_bytes))
```
ronaldmaj, line 36: I was able to test this and works well

gnssanalysis/gn_io/sinex.py
```diff
@@ -25,17 +25,21 @@
...
     header_end = snx_bytes.find(b"\n", header_begin)
+    if header_end == -1:
+        raise ValueError("Sinex header found but wasn't newline terminated (end unclear)", file_desc(path_or_bytes))
```
ronaldmaj, line 39: I've tried to end up here but wasn't successful in editing an SNX to do so - do you have an example file where you were able to?

gnssanalysis/gn_io/sinex.py
```diff
@@ -25,17 +25,21 @@
...
     snx_hline = snx_bytes[header_begin:header_end].decode()
     if len(snx_hline) == 0: # Raise an inteligible error if the Sinex header is effectively empty.
-        # Is the source bytes, or a path / str path? If not bytes, we can output it.
-        if isinstance(path_or_bytes, bytes):
-            raise ValueError(f"Sinex header is empty for UNKNOWN file (passed in as bytes object not path)")
-        else:
-            raise ValueError(f"Sinex header is empty for file: '{str(path_or_bytes)}'")
+        raise ValueError(f"Sinex header found, but empty!", file_desc(path_or_bytes))
```
ronaldmaj, line 42: Again, wasn't able to get here.
I've tried just having the `%=SNX` with nothing afterwards, but the error is around pandas not finding columns I believe:
```
EmptyDataError: No columns to parse from file
```
Anyway, none of this is a show stopper so I think it can go in

ronaldmaj (approved):
Tested this, and works for normal SNX file, and with ones missing header info - just wasn't able to get all different error cases.
I'm happy for this to go in though

PR #29, opened 2024-07-04
HerlmertPrint

seballgeyer:
Issue raised by Simon. 
The information from the helmert parameters are not complete, only the first parameter is printed. 
This PR change so all the vector of parameter is printed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,5 +10,6 @@ __pycache__
 
 # IDE specific details
 .vscode
-
+.idea
 # Other files
+scratch
--- a/gnssanalysis/gn_utils.py
+++ b/gnssanalysis/gn_utils.py
@@ -565,7 +565,7 @@ def orbq(input, output_path, format, csv_separation, json_format, hlm_mode, sate
     rms_df = gn_diffaux.rac_df_to_rms_df(rac)
 
     if hlm_mode is not None:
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -10,5 +10,6 @@ __pycache__
 
 # IDE specific details
 .vscode
-
-# Other files
\ No newline at end of file
+.idea
```
ronaldmaj, line 13: Is this a new directory / file type?

gnssanalysis/gn_utils.py
```diff
@@ -565,7 +565,7 @@ def orbq(input, output_path, format, csv_separation, json_format, hlm_mode, sate
     rms_df = gn_diffaux.rac_df_to_rms_df(rac)
 
     if hlm_mode is not None:
-        print(f"Helmert coeffs computed in {hlm_mode}: {rac.attrs['hlm'][0][0].reshape(-1)}")
+        print(f"Helmert coeffs computed in {hlm_mode}: {rac.attrs['hlm'][0].reshape(-1)}")
```
ronaldmaj, line 568: Seems like a simple enough change, no issue with this going in

ronaldmaj (approved):
I think this is a pretty simple change, happy for this to go in and it'll help display all info that is calculated

seballgeyer (commented):
> Is this a new directory / file type? 

it's the directory, where pycharm store all its config (like .vscode for vscode)

PR #34, opened 2024-07-26
modify to add capability to read both Ginan and Bernese troposphere .…

umma-zannat:
to support both Ginan and Bernese tropospheric files (`.tro` files)


ronaldmaj (commented):
Looks like some good changes coming in here to expand `.tro` reading 
- Do you have example `.tro` files we can test the new function on?

umma-zannat:
The other format for troposhere is `cost` file. `.tro` file is basically a sinex formatted file.

umma-zannat:
`    trop_sinex:                                 # Troposphere SINEX files are used to export troposhere products, such as ZTD and delay gradients
        output: true                           # Enable data exporting to troposphere SINEX file
        const_code:                                     # Troposphere SINEX const code
        directory:      "<OUTPUTS_ROOT>"                # Directory to export troposphere SINEX file
        filename:       "<CONFIG>-<LOGTIME>.tro" # Troposphere SINEX filename
        obs_code:       P                               # Troposphere SINEX observation code
        sol_type:       "Solution parameters"           # Troposphere SINEX solution type
        sources:        [KALMAN]                        # Source for troposphere delay data - KALMAN, etc
        version:        2                               # Troposphere SINEX version

    # log:`

gnssanalysis/gn_io/trop.py
```diff
@@ -9,41 +9,70 @@
 from .. import gn_io as _gn_io
 
 
-def _read_tro_solution(path: str, recenter: bool = True) -> _pd.DataFrame:
+def _read_tro_solution(path: str, recenter: bool = True, trop_mode="Ginan") -> _pd.DataFrame:
+    """For backwards compatibility"""
```
ronaldmaj, line 13: Docstring should be one-line sphinx (missing the description of input parameters, errors raise, what's returned, etc.: `:param` `:return`)
https://github.com/NilsJPWerner/autoDocstring/blob/master/docs/one-line-sphinx.md

gnssanalysis/gn_io/trop.py
```diff
@@ -9,41 +9,70 @@
...
+    return read_tro_solution(path, recenter=recenter, trop_mode=trop_mode)
+
+
+def read_tro_solution(path: str, recenter: bool = True, trop_mode="Ginan") -> _pd.DataFrame:
     """Parses tro snx file into a dataframe.
```
ronaldmaj, line 18: One-line sphinx docstring (missing the description of input parameters, errors raise, what's returned, etc.: `:param` `:return`)

ronaldmaj (changes requested):
Again, looks good, couple docstrings that need to be finished off, but I think once you've got the unittests that Seb suggested in, this should be good to go

ronaldmaj (commented):
Thanks heaps for putting through the `docstring` additions :pray: 

I've noticed you now have two `read_tro_solution` functions - is this intentional?
There is `_read_tro_solution` and `read_tro_solution`

umma-zannat:
`If no one uses the function with underscore I can delete it. I personally prefer not to use underscore. `

ronaldmaj (approved):
Unit tests are a great addition - can use that instead of testing manually. Happy for this to go in :+1:

gnssanalysis/gn_io/trop.py
ronaldmaj: This function is up for deletion if no one uses it (as there is the `read_tro_solution` function below this

PR #35, opened 2024-07-29
Npi 3421 Fix nodata formatting of standard deviation columns

treefern:
Column names specified for formatter mapping were subtly incorrect. This prevented the relevant formatting function from being invoked.

Also added: minor doc improvements to clarifying nodata handling logic for POS columns.


ronaldmaj (approved):
Thanks for putting this through. We've discussed this previously and the inclusion of the `_` in the `STD` str for the formatters enables the value to actually be read and outputted.
Good for this to go in :+1:

Issue #28, opened 2024-06-19
Cannot merge sp3 files with sp3merge properly


different91988:
Two SP3 files download from IGS AC cannot be merged properly. 
The file urls are followings:
ftp://igs.gnsswhu.cn/pub/gps/products/2286/WUM0MGXFIN_20233020000_01D_05M_ORB.SP3.gz
ftp://igs.gnsswhu.cn/pub/gps/products/2286/USN0OPSULT_20233040000_02D_15M_ORB.SP3.gz
Same problem occured when try to merge other IGS MGEX SP3 files. The exception throws when try to merge two dataframes read from sp3 files in Line 469 of file sp3.py eg. merged_sp3 = _pd.concat(sp3_dfs)

Traceback (most recent call last):
  File "D:\Program Files\JetBrains\PyCharm 2024.1.3\plugins\python\helpers\pydev\pydevd.py", line 1546, in _exec
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "D:\Program Files\JetBrains\PyCharm 2024.1.3\plugins\python\helpers\pydev\_pydev_imps\_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "D:\pythonProj\gnssanalysis\tests\test_diff.py", line 8, in <module>
    merged_df = _gn_io.sp3.sp3merge(["D:\\MATLAB\\compareSp3\\code\\WUM0MGXFIN_20232020000_01D_05M_ORB.SP3",
  File "D:\pythonProj\gnssanalysis\gnssanalysis\gn_io\sp3.py", line 469, in sp3merge
    merged_sp3 = _pd.concat(sp3_dfs)
  File "D:\Anaconda\envs\myenv\lib\site-packages\pandas\core\reshape\concat.py", line 385, in concat
    return op.get_result()
  File "D:\Anaconda\envs\myenv\lib\site-packages\pandas\core\reshape\concat.py", line 623, in get_result
    return cons(new_data).__finalize__(self, method="concat")
  File "D:\Anaconda\envs\myenv\lib\site-packages\pandas\core\generic.py", line 5963, in __finalize__
    check_attrs = all(objs.attrs == attrs for objs in other.objs[1:])
  File "D:\Anaconda\envs\myenv\lib\site-packages\pandas\core\generic.py", line 5963, in <genexpr>
    check_attrs = all(objs.attrs == attrs for objs in other.objs[1:])
[...]

PR #37, opened 2024-07-30
NPI-3429 - Fix mergesp3 utility 

ronaldmaj:
This PR is created to address the issue brought up here: 
https://github.com/GeoscienceAustralia/gnssanalysis/issues/28

Basically the `mergesp3` utility in it's current form was not doing what it was supposed to.
Giving it a pair of any two sp3 files led to an error each time (a `ValueError` saying that the truth value of two series was ambiguous). This was coming up because in using the `pd.concat` function in `gn_io.sp3.sp3merge` there is a step where the equality of the `df.attrs` are checked. These are going to be different for sp3 files from different time periods.

Therefore, I've created a new variable that merges the `attrs` separately
```py
merged_attrs = merge_attrs(sp3_dfs)
```
using an existing function `merge_attrs`, and then assigns the `attrs` of each `df` to an empty `dict` (thereby making them the same).
```py
    for df in sp3_dfs:
        df.attrs = {}
```

One problem I encoutnered that took a while to debug was that even with no CLK files, I would keep ending up in the block of code that tries to incorporate CLK files into the dataframe, causing failure.
This is due to the `mergesp3` utility producing an empty tuple rather than a `None` when no clkpaths are passed in.

I've manually set it to `None` now in that case.


gnssanalysis/gn_io/sp3.py
```diff
@@ -579,13 +579,16 @@ def sp3merge(
     :return pd.DataFrame: The merged sp3 DataFrame.
     """
     sp3_dfs = [read_sp3(sp3_file, nodata_to_nan=nodata_to_nan) for sp3_file in sp3paths]
+    # Create a new attrs dictionary to be used for the output DataFrame
+    merged_attrs = merge_attrs(sp3_dfs)
```
seballgeyer, line 583: It looks like to me that the merge_attrs is quite simplistic. 
if values in the header are different, replace most of the values by "MIX". 
Maybe a better strategy, 
start date = min(start dates)
n_epochs = sum(n_epochs)
....

seballgeyer (commented):
It would be interesting also to have some unit test on this one, just making sure that the output is what we expect.

ronaldmaj:
> It would be interesting also to have some unit test on this one, just making sure that the output is what we expect.

Yeah, good point, forgot about this. I'll put together something now - I need to get into the habit of writing unit tests when I do work on `gnssanalysis` :smile:

gnssanalysis/gn_io/sp3.py
ronaldmaj: Yeah I didn't look too much into that `merge_attrs` function - seems like it was an MVP then.
Not sure for some of these header fields, but `DATETIME` and `N_EPOCHS` I can definitely be write out by analysing the data of the incoming DataFrames.

An example block is given below:
```
VERSION                                 c
PV_FLAG                                 P
DATETIME     2023 10 29  0  0  0.00000000
N_EPOCHS                              288
DATA_USED                             u+U
COORD_SYS                           IGS20
ORB_TYPE                              FIT
AC                                    GFZ
FILE_TYPE                               M
TIME_SYS                              GPS
```
Not sure if there are any other fields where `MIX` is not appropriate?

ronaldmaj: I could perhaps label some fields with just a new string that incorporates the ones from the original files
- e.g. for `VERSION` we could label as `c+d` when the two files are Version `c` and `d`
- or the AC could be `GFZ+WHU` when combining those two
It won't tell us which part of the data comes from each, but at least gives the user an idea of what is in it

ronaldmaj: Ok, just pushed through a commit that allows the DATETIME and N_EPOCHS fields to be filled properly - based on the data in the new merged file. 
For some of the other fields, I've had to use a certain number of characters, so I couldn't achieve my aim of just appending each instance with a `+`, although I've tried to get as close to that aim as possible (e.g. `AC` uses all 4 chars available and combines 2 chars from each AC)

Also, the way this function originally outputted the first line, it couldn't be re-read by our `read_sp3` function. I think I've got it working now so that we can read the outputs again

ronaldmaj:
> It would be interesting also to have some unit test on this one, just making sure that the output is what we expect.

Okie doke, I've been able to write up a test for the `sp3merge` function. I did have to introduce a new py package unfortunately (`pyfakefs`) but that was the only way I saw to mimic a filesystem and therefore test the function itself.

Test seems to pass the pipeline ok:

seballgeyer (approved):
all good,
tested, pushed one correction wrt numpy v.2 compatibility

seballgeyer (approved):

gnssanalysis/gn_io/sp3.py
```diff
@@ -508,9 +508,9 @@ def clk_std_formatter(x):
 
         # POS nodata formatting
         # Fill +/- infinity values with SP3 nodata value for POS columns
-        epoch_vals["X"].replace(to_replace=[_np.inf, _np.NINF], value=SP3_POS_NODATA_STRING, inplace=True)
-        epoch_vals["Y"].replace(to_replace=[_np.inf, _np.NINF], value=SP3_POS_NODATA_STRING, inplace=True)
-        epoch_vals["Z"].replace(to_replace=[_np.inf, _np.NINF], value=SP3_POS_NODATA_STRING, inplace=True)
+        epoch_vals["X"].replace(to_replace=[_np.inf, -_np.inf], value=SP3_POS_NODATA_STRING, inplace=True)
```
treefern, line 511: This will need updating again for Pandas 3. I'll try to replace it with `format()` calls at the point of writing the data out.

ronaldmaj: > This will need updating again for Pandas 3. I'll try to replace it with `format()` calls at the point of writing the data out.

Yeah, there's a few changes that we'll need to make in order to make us Pandas 3 compliant. Running the `sp3merge` utility already has a whole bunch of warnings around this

tests/test_sp3.py
```diff
@@ -89,3 +187,29 @@ def test_velinterpolation(self, mock_file):
...
+        epoch_index = result.index.get_level_values("J2000")
+        sat_index = result.index.get_level_values("PRN")
+        # Verify
+        self.assertEqual(min(epoch_index), 229608000)
+        self.assertEqual(max(epoch_index), 774619500)
+        self.assertEqual(sat_index[0], "G01")
+        self.assertEqual(sat_index[-1], "R02")
+        self.assertEqual(result.attrs["HEADER"].HEAD.VERSION, "d")
+        self.assertEqual(result.attrs["HEADER"].HEAD.AC, "AIES")
+        self.assertEqual(result.attrs["HEADER"].HEAD.COORD_SYS, None)
```
treefern, line 215: Might be worth adding another one ensuring that the merged file ended up with PV_FLAG=`P`
(GitHub doesn't seem to have a way of writing a suggestion that adds a line, without modifying an existing one)
```suggestion
        self.assertEqual(result.attrs["HEADER"].HEAD.COORD_SYS, None)
        self.assertEqual(result.attrs["HEADER"].HEAD.PV_FLAG, "P")
```

treefern (approved):
Looks good to me.
I examined the padded date-stamp modifying line and couldn't think of a way it could go wrong.
It's nice to see some more unit tests coming in.

treefern (approved):

PR #38, opened 2024-08-02
Npi 3430 Fix SP3 header parsing to more robustly handle orbit accuracy values

treefern:
Previous implementation searched for a slightly incorrect length of data (50 instead of 51 chars), and did this instead of explicitly specifying the expected format (three char chunks which always end with a digit). It also did not allow for `-`, leading negative values to crash the SP3 header parser.

### DIFF
--- a/gnssanalysis/gn_io/sp3.py
+++ b/gnssanalysis/gn_io/sp3.py
@@ -25,10 +25,13 @@ _RE_SP3_HEAD = _re.compile(
                                     (\w+)(?:[ ]+(\w+)|)""",
     _re.VERBOSE,
 )
-# SV names. multiline, findall
+# Regex to extract Satellite Vehicle (SV) names (E.g. G02). Regex options/flags: multiline, findall
 _RE_SP3_HEAD_SV = _re.compile(rb"^\+[ ]+(?:[\d]+|)[ ]+((?:[A-Z]\d{2})+)\W", _re.MULTILINE)
-# orbits accuracy codes
-_RE_SP3_ACC = _re.compile(rb"^\+{2}[ ]+([\d\s]{50}\d)\W", _re.MULTILINE)
+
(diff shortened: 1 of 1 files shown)


treefern:
Further commentary, more minor fixes, and unit tests to be added in another PR.

ronaldmaj (approved):
This looks good, change is mostly around fixing a regex that impacts the `read_sp3` function. With the pipeline passing, and therefore the unit-test for `read_sp3` passing too, happy to push this through

gnssanalysis/gn_io/sp3.py
```diff
@@ -25,10 +25,25 @@
...
 )
-# SV names. multiline, findall
+# Regex to extract Satellite Vehicle (SV) names (E.g. G02). Regex options/flags: multiline, findall
 _RE_SP3_HEAD_SV = _re.compile(rb"^\+[ ]+(?:[\d]+|)[ ]+((?:[A-Z]\d{2})+)\W", _re.MULTILINE)
-# orbits accuracy codes
-_RE_SP3_ACC = _re.compile(rb"^\+{2}[ ]+([\d\s]{50}\d)\W", _re.MULTILINE)
+
+# Regex for orbit accuracy codes (E.g. ' 15' - space padded, blocks are three chars wide).
+# Note: header is padded with '  0' entries after the actual data, so empty fields are matched and then trimmed.
+_RE_SP3_HEADER_ACC = _re.compile(rb"^\+{2}[ ]+((?:[\-\d\s]{2}\d){17})$", _re.MULTILINE)
```
treefern, line 33: It turns out the use of the trailing `$` here was a mistake. Not all SP3 files have the line ending immediately after the orbit accuracy codes. Some pad with spaces. This is resolved in https://github.com/GeoscienceAustralia/gnssanalysis/pull/45

PR #39, opened 2024-08-02
NPI-3430 Improve SP3 parsing robustness

treefern:
- Implemented pre-parse SP3 **comment line stripping** and updated header end detection to use a more robust anchor.
- Implemented check that the **number of SV entries the header says are there**, is the number we get, and warning if it is not.
- Added **warning for negative orbit accuracy** values (as these are extremely unrealistic and probably an error).
- Additional SP3 parsing unit tests, including:
  - SVs: correct number, no empty names.
  - Orbit accuracy values, including **correct parsing of negative values** (however implausible they are in valid data).
    - **This should address the internal pipeline error reported in NPI-3430.**
  - Correct extraction of first epoch value as both a string and J2000.
    - This serves as a way of checking that the header end detection did not overshoot and incorrectly mark part of the data section as the header.
- Various formatting fixes to make (primarily the new code) more closely match `Black` requirements.


ronaldmaj (approved):
Thanks for going through and getting these changes in :pray:  (plus the unit-tests)

The one comment I have on this is that it looks like you've got your python `black` formatter set to the usual 80 char per line, we use 120 char per line as standard. 

I think this is still good to put through as functionally it's good, but we'll need to change it with another PR (perhaps a general formatting one, and just run `black` over all files).

PR #40, opened 2024-08-03
fix: Hopefully fix the publish papeline, wrong file name.

seballgeyer:
Trying to fix the publish pipeline. 
Issue for not publishing was due to the failure in to call the unit test pipeline properly (filename python-cicd-units.yml vs  python-cicd-units.yaml)


ronaldmaj (approved):
Fair enough, happy for this to go in

PR #41, opened 2024-08-04
update: publish workflow (on publish only)

seballgeyer:
Updating the publish workflow. 
The tag and previous CICD pipeline dependency have been removed, as the publish workflow is runing everytime after the CICD but skipped if not a release. 
It will be to the user to make sure the CICD workflow is successful before creating a release.


ronaldmaj (approved):
Looks good to me - this will ensure we can get `gnssanalysis` published to PyPI after each release via the pipeline again. Happy for this to go in :+1:

PR #42, opened 2024-08-05
Pipeline-fix

seballgeyer:
Now it should be good. Issues was in the non-copy of the requirements.txt; now it is specified in the MANIFEST file


ronaldmaj (approved):
No issues here

PR #43, opened 2024-08-05
Npi 3294 Fixes and better warnings in filename based property extraction, and filename vs contents checks

treefern:
Various improvements and fixes to functionality initially introduced in https://github.com/GeoscienceAustralia/gnssanalysis/pull/21


ronaldmaj (approved):
A good set of checks and warnings for filenames based on the contents of the data, including unit-tests for the new functions

PR #44, opened 2024-08-12
ZTD cpmparison

umma-zannat:
Ginan outputs to total delay in ztd in every 20 seconds whereas Bernese does it in every one hour.  So here, I am averaging Ginan data over an hour to match it with Bernese ultra-rapid products. Converting to python datetime makes it easier for pandas dataframe.


seballgeyer (approved):
Looks good to me, pretty straightforward changes. 
I like the unittest looping ;)

PR #45, opened 2024-08-13
NPI-3446 Fix SP3 parse and write of data section flags, plus address CLK and POS nodata causing misalignment

treefern:
- fixed incorrect width of unused columns between data flag values (2 instead of 1). This allows the second prediction flag to be read in.
- Consolidated definitions of columns for maintainability.
- Updated column/index definitions to clearly define a FLAGS index, and specify names of individual flag columns.
- Fixed SP3 output generator to access and output FLAGS data.
- Addressed misalignment bug caused by the combination of CLK nodata value width, and Pandas adding a space between columns. Also applied this fix to POS nodata value, as it appears the same issue would apply there.
- Added handling for missing flag values (conformance to empty strings to avoid outputting nans)
- Added placeholder TODOs for throwing exceptions if record types are encountered that we currently do not implement support for (EP, EV).

### DIFF
--- a/gnssanalysis/gn_io/sp3.py
+++ b/gnssanalysis/gn_io/sp3.py
@@ -691,7 +691,7 @@ def gen_sp3_content(
         "STD_Z": pos_std_formatter,
         "STD_CLK": clk_std_formatter,  # ditto above
     }
-    for epoch, epoch_vals in out_df.reset_index("PRN").groupby(axis=0, level="J2000"):
+    for epoch, epoch_vals in out_df.reset_index("PRN").groupby(level="J2000"):
         # Format and write out the epoch in the SP3 format
         epoch_datetime = _gn_datetime.j2000_to_pydatetime(epoch)
         frac_seconds = epoch_datetime.second + (epoch_datetime.microsecond * 1e-6)
--- a/gnssanalysis/gn_io/sp3.py
+++ b/gnssanalysis/gn_io/sp3.py
@@ -35,7 +35,7 @@ _RE_SP3_HEADER_SV = _re.compile(
 
 # Regex for orbit accuracy codes (E.g. ' 15' - space padded, blocks are three chars wide).
 # Note: header is padded with '  0' entries after the actual data, so empty fields are matched and then trimmed.
-_RE_SP3_HEADER_ACC = _re.compile(rb"^\+{2}[ ]+((?:[\-\d\s]{2}\d){17})$", _re.MULTILINE)
+_RE_SP3_HEADER_ACC = _re.compile(rb"^\+{2}[ ]+((?:[\-\d\s]{2}\d){17})\W", _re.MULTILINE)
 # This matches orbit accuracy codes which are three chars long, left padded with spaces, and always contain at
 # least one digit. They can be negative, though such values are unrealistic. Empty 'entries' are '0', so we have to
 # work out where to trim the data based on the number of SVs in the header.
@@ -43,11 +43,12 @@ _RE_SP3_HEADER_ACC = _re.compile(rb"^\+{2}[ ]+((?:[\-\d\s]{2}\d){17})$", _re.MUL
 # - Match the accuracy code line start of '++' then arbitary spaces, then
 # - 17 columns of:
 #   [space or digit or - ] (times two), then [digit]
-# - Then end of line.
+# - Then non-word char, to match end of line / space.
 
-# Most data matches whether or not we specify the end of line, as we are quite explicit about the format of
-# data we expect. However, using \W instead can be risky, as it fails to match the last line if there is no
-# newline following it. Note though that in SV parsing $ doesn't traverse multiline, the \W works...
+# NOTE: looking for the *end of the line* (i.e. with '$') immediately following the data, would fail to handle files
+# which pad the remainder with spaces.
+# Using \W fails to match the last line if there isn't a newline following, but given the next line should be the %c
+# line, this should never be an issue.
 
 # File descriptor and clock
 _RE_SP3_HEAD_FDESCR = _re.compile(rb"\%c[ ]+(\w{1})[ ]+cc[ ](\w{3})")
--- a/gnssanalysis/gn_io/sp3.py
+++ b/gnssanalysis/gn_io/sp3.py
@@ -146,7 +146,8 @@ SP3_VELOCITY_COLUMNS = [
 # on adding a space between columns (so this cancels out the missing space here).
 # For the POS value, no leading spaces are required by the SP3 spec, but we need the total width to be 13 chars,
 # not 14 (the official width of the column i.e. F14.6), again because Pandas insists on adding a further space.
-# See comment in gen_sp3_content() line ~622 for further discussion.
+# See comment in gen_sp3_content() line ~645 for further discussion.
+# Another related 'hack' can be found at line ~602, handling the FLAGS columns.
 SP3_CLOCK_NODATA_STRING = "999999.999999"
 SP3_CLOCK_NODATA_NUMERIC = 999999
 SP3_POS_NODATA_STRING = "     0.000000"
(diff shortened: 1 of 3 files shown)


ronaldmaj:
Running this branch in python 3.12 via `ipython` I received the following warning:
```
In [2]: from gnssanalysis.gn_io.sp3 import read_sp3, write_sp3
/data/review-repos/gnssanalysis/gnssanalysis/gn_io/blq.py:12: SyntaxWarning: invalid escape sequence '\/'
  _RE_LLH = _re.compile(b"lon\/lat\:\s+([\d\.]+)+\s+([\d\.\-]+)\s+([\d\.\-]+)")
/data/review-repos/gnssanalysis/gnssanalysis/gn_io/trop.py:76: SyntaxWarning: invalid escape sequence '\s'
  sep='\s+',
```
Not sure if we need to resolve these here, but something to take note of

gnssanalysis/gn_io/sp3.py
```diff
@@ -53,7 +53,7 @@
 _RE_SP3_HEAD_FDESCR = _re.compile(rb"\%c[ ]+(\w{1})[ ]+cc[ ](\w{3})")
 
 
-_SP3_DEF_PV_WIDTH = [1, 3, 14, 14, 14, 14, 1, 2, 1, 2, 1, 2, 1, 3, 1, 1, 1, 1, 1, 1]
+_SP3_DEF_PV_WIDTH = [1, 3, 14, 14, 14, 14, 1, 2, 1, 2, 1, 2, 1, 3, 1, 1, 1, 2, 1, 1]
```
ronaldmaj, line 56: This probably isn't where the problem is arising, but seems like we are getting a misalignment of the output `P` column.

I've run a couple tests of SP3 files, reading and then writing them back out, and comparing the input and output file, there is an added space somewhere causing the `P` column to be shifted by one across:

Original IGS:
```
PG31  -6665.022843 -21628.559592  13360.498756   -226.172552  6  6  6 161       
PG32  15155.161510 -15052.638054  15871.232492   -608.029872  8  7  7 143       
*  2024  8  5  0  0  0.00000000
PG02 -22682.219867 -13324.112236   5209.362171   -385.403423  6  8  9 167  P   P
PG03 -13555.871391  -5841.138951  21922.234449    487.153854  9  6  7 207  P   P
```
After read / write cycle:
```
PG31  -6665.022843 -21628.559592  13360.498756   -226.172552  8  8  8 206        
PG32  15155.161510 -15052.638054  15871.232492   -608.029872  9  9  9 201        
*  2024  8  5  0  0  0.00000000
PG02 -22682.219867 -13324.112236   5209.362171   -385.403423  8  9 10 207   P   P
PG03 -13555.871391  -5841.138951  21922.234449    487.153854 10  8  9 216   P   P
```
An unrelated issue revealed by this is also that the STD values are being changed, but we can deal with that in another PR

gnssanalysis/gn_io/sp3.py
ronaldmaj, line 1: Running a read / write test for a GFZ file seems to throw up an error:
```py
In [27]: E = read_sp3("GFZ0OPSRAP_20233020000_01D_05M_ORB.SP3")
ValueError                                Traceback (most recent call last)
Cell In[27], line 1
----> 1 E = read_sp3("GFZ0OPSRAP_20233020000_01D_05M_ORB.SP3")

File /data/review-repos/gnssanalysis/gnssanalysis/gn_io/sp3.py:278, in read_sp3(sp3_path, pOnly, nodata_to_nan)
    276 header = content[:header_end]
    277 content = content[header_end:]
--> 278 parsed_header = parse_sp3_header(header)
    279 counts = parsed_header.SV_INFO.count()
    280 fline_b = header.find(b"%f") + 2  # TODO add to header parser

File /data/review-repos/gnssanalysis/gnssanalysis/gn_io/sp3.py:432, in parse_sp3_header(header, warn_on_negative_sv_acc_values)
    415     logger.warning(
    416         "Number of Satellite Vehicle (SV) entries extracted from the SP3 header, did not match the "
    417         "number of SVs the header said were there! This might be a header writer or header parser bug! "
    418         f"SVs extracted: {found_sv_count}, SV count given by header: {head_sv_expected_count} "
    419         f"List of SVs extracted: '{str(head_svs)}'"
    420     )
    422 # Use regex to extract the Orbit Accuracy Codes from the header. These correspond with the above
    423 # Satellite Vehicle (SV) numbers. Values are left padded and each takes up three characters. E.g. ' 15'.
    424 # The data structure itself is zero padded and spread over multiple lines. This comes out as e.g.:
   (...)
    428 # Note: .view("S3") seems to present the data in chunks of three characters (I'm inferring this
    429 # though, doco is unclear).
    430 head_svs_std = (
    431     _np.asarray(b"".join(_RE_SP3_HEADER_ACC.findall(header)))[_np.newaxis]
--> 432     .view("S3")[: head_svs.shape[0]]
    433     .astype(int)
    434 )
    435 sv_tbl = _pd.Series(head_svs_std, index=head_svs)
    437 if warn_on_negative_sv_acc_values and any(acc < 0 for acc in head_svs_std):

ValueError: When changing to a larger dtype, its size must be a divisor of the total size in bytes of the last axis of the array.
```
Here is the file in question: 
[GFZ0OPSRAP_20233020000_01D_05M_ORB.SP3.tar.gz](https://github.com/user-attachments/files/16607205/GFZ0OPSRAP_20233020000_01D_05M_ORB.SP3.tar.gz)

This file was working previously as I used it to test out the `sp3_merge` utility

ronaldmaj (changes requested):
Thanks for putting this PR together. It hits a lot of the issues we've been seeing recently in the Ginan-Ops space.

I found some errors in doing reading / writing cycles with example SP3 files. I think when these are resolved, this should be good to go in

gnssanalysis/gn_io/sp3.py
treefern: I've revised the way the `FLAGS` columns are written out, which addresses this.
Flags are now concatenated into a single string per record in the dataframe, ahead of the call to `DataFrame.to_string()`. This works around the compulsory space inserted between columns, which was causing misalignment.

gnssanalysis/gn_io/sp3.py
treefern: This turns out to be a bug I introduced in https://github.com/GeoscienceAustralia/gnssanalysis/pull/38 

That regex assumed orbit accuracy code lines would end with a newline. That isn't the case here, where the end of the line is space padded.

Given the specification dictates that orbit accuracy lines will end at column 60, and we explicitly match the 17 entries, there is no need to match the end of the line (or a space at the end). For consistency, we'll match a non-word character as was done previously.

This should now be fixed.

ronaldmaj (dismissed):
Tested with various SP3 files from different ACs (COD, GFZ, WHU, GA) and different types (ULT, RAP, FIN). Seems to work great now!
Thanks for putting through those changes yesterday. 
I think this now good to go in :+1:

treefern:
> Running this branch in python 3.12 via `ipython` I received the following warnings:
> 
> ```python
> In [2]: from gnssanalysis.gn_io.sp3 import read_sp3, write_sp3
> /data/review-repos/gnssanalysis/gnssanalysis/gn_io/blq.py:12: SyntaxWarning: invalid escape sequence '\/'
>   _RE_LLH = _re.compile(b"lon\/lat\:\s+([\d\.]+)+\s+([\d\.\-]+)\s+([\d\.\-]+)")
> /data/review-repos/gnssanalysis/gnssanalysis/gn_io/trop.py:76: SyntaxWarning: invalid escape sequence '\s'
>   sep='\s+',
> ```
> 
> Not sure if we need to resolve these here, but something to take note of

Resolved with use of `raw` string flag (i.e. `'rb"blah"`). Testing on another example, it seems it should work either way, but this makes the linter happy.

ronaldmaj (approved):
Additional commits to address comment I made around the regex expression and leaving notes - minor changes that don't change the overall SP3 changes so happy for this to go in

PR #46, opened 2024-08-16
NPI-3456 Transfer auto_download_PPP functions

ronaldmaj:
Relatively quick PR to transfer some useful functions from the `auto_download_PPP` script in the Ginan repo into `gnssanalysis`

I've also removed some deprecated functions at the same time.

Functions moved across: `ensure_folders`, `configure_logging`, `download_product_from_cddis`, `download_atx`, `download_satellite_metadata_snx`, `download_yaw_files`


seballgeyer (commented):
Looks good. 
We might need to find a way to reduce the file of gn_download.py (1,025 lines now). in the near future.

gnssanalysis/gn_download.py
```diff
@@ -846,325 +848,169 @@ def download_multiple_files_from_cddis(files: [str], ftp_folder: str, output_fol
...
+            if not (product_filename in all_files):
+                logging.info(f"{product_filename} not in gnss/products/{gps_date.gpswk} - too recent")
+                raise FileNotFoundError
+
+        # reference_start will be changed in the first run through while loop below
+        reference_start -= _datetime.timedelta(hours=24)
+        count = 0
+        remain = end_epoch - reference_start
+        while remain.total_seconds() > timespan.total_seconds():
+            if count == limit:
```
treefern, line 922: It might be clearer to express this as
```suggestion
            if iteration_count == days_limit:
                break
```

gnssanalysis/gn_download.py
```diff
@@ -846,325 +848,169 @@ def download_multiple_files_from_cddis(files: [str], ftp_folder: str, output_fol
...
+            )
+            ftps.cwd(f"gnss/products/{gps_date.gpswk}")
+
+            all_files = ftps.nlst()
+            if not (product_filename in all_files):
+                logging.info(f"{product_filename} not in gnss/products/{gps_date.gpswk} - too recent")
+                raise FileNotFoundError
+
+        # reference_start will be changed in the first run through while loop below
+        reference_start -= _datetime.timedelta(hours=24)
```
treefern, line 918: This acts like a shift interval. `reference_start` implies that it refers to the start epoch in some way.
```suggestion
        shift_interval -= _datetime.timedelta(hours=24)
```

gnssanalysis/gn_download.py
```diff
@@ -846,325 +848,169 @@ def download_multiple_files_from_cddis(files: [str], ftp_folder: str, output_fol
...
+    logging.debug("Attempting CDDIS Product download/s")
+    logging.debug(f"Start Epoch - {start_epoch}")
+    logging.debug(f"End Epoch - {end_epoch}")
+
+    reference_start = _deepcopy(start_epoch)
+    product_filename, gps_date, reference_start = generate_product_filename(
+        reference_start,
+        file_ext,
+        long_filename=long_filename,
+        AC=analysis_center,
```
treefern, line 884: Not sure what happened here
```suggestion
        analysis_center=analysis_center,
```

gnssanalysis/gn_download.py
```diff
@@ -846,325 +848,169 @@ def download_multiple_files_from_cddis(files: [str], ftp_folder: str, output_fol
...
+            ftps.cwd(f"gnss/products/{gps_date.gpswk}")
+        except _ftplib.all_errors as e:
+            logging.info(f"{reference_start} too recent")
+            logging.info(f"ftp_lib error: {e}")
+            product_filename, gps_date, reference_start = generate_product_filename(
+                reference_start,
+                file_ext,
+                shift=-6,
+                long_filename=long_filename,
+                AC=analysis_center,
```
treefern, line 904: As above
```suggestion
        analysis_center=analysis_center,
```

gnssanalysis/gn_download.py
```diff
@@ -846,325 +848,169 @@ def download_multiple_files_from_cddis(files: [str], ftp_folder: str, output_fol
...
+        while remain.total_seconds() > timespan.total_seconds():
+            if count == limit:
+                remain = _datetime.timedelta(days=0)
+            else:
+                product_filename, gps_date, reference_start = generate_product_filename(
+                    reference_start,
+                    file_ext,
+                    shift=24,  # Shift at the start of the loop - speeds up total download time
+                    long_filename=long_filename,
+                    AC=analysis_center,
```
treefern, line 930: And again
```suggestion
        analysis_center=analysis_center,
```

gnssanalysis/gn_download.py
ronaldmaj: Yeah, so the `reference_start` is effectively a pointer used to determine where to start looking for files / downloading files.
This line shifts that `reference_start` back 24 hours, but the variable is returned: `-=` operation acts on the variable and returns it.
This is effectively: `reference_start = reference_start - _datetime.timedelta(hours=24)`

gnssanalysis/gn_download.py
```diff
@@ -846,325 +848,169 @@ def download_multiple_files_from_cddis(files: [str], ftp_folder: str, output_fol
...
+    return download_filepath
 
-    if isinstance(stations, str):
-        stations = [stations]
 
-    # Create directory if doesn't exist:
-    if not _Path(dest).is_dir():
-        _Path(dest).mkdir(parents=True)
+def download_yaw_files(download_dir: _Path, if_file_present: str = "prompt_user") -> _List[_Path]:
+    """_summary_
```
treefern, line 986: This looks like a placeholder TODO ?

gnssanalysis/gn_download.py
```diff
@@ -846,325 +848,169 @@ def download_multiple_files_from_cddis(files: [str], ftp_folder: str, output_fol
...
 
-    if f_dict:
-        f_dict = {"rnxfiles": []}
+    :param _Path download_dir: Directory to store the file in
+    :param str if_file_present: What to do if file already present: "replace", "dont_replace", defaults to "prompt_user"
+    :return _List[_Path]: _description_
+    """
+    ensure_folders([download_dir])
+    urls = [
+        "https://peanpod.s3.ap-southeast-2.amazonaws.com/aux/products/tables/bds_yaw_modes.snx.gz",
```
treefern, line 994: We should potentially be constructing this from a config class or something, rather than hard-coding the full value.

gnssanalysis/gn_download.py
treefern: Oops, not sure how I missed the `-` in that line.

gnssanalysis/gn_download.py
ronaldmaj: This counts the number of files that have been downloaded. If it surpasses the limit, loop exists and downloading stops

ronaldmaj:
> Looks good. We might need to find a way to reduce the file of gn_download.py (1,025 lines now). in the near future.

Agreed, definitely getting a little too long now

gnssanalysis/gn_download.py
```diff
@@ -854,17 +854,29 @@ def download_product_from_cddis(
...
-    Download the file/s from CDDIS based on start and end epoch, to the
-    provided the download directory (download_dir)
+    """Download the file/s from CDDIS based on start and end epoch, to the download directory (download_dir)
+
+    :param _Path download_dir: Where to download files
+    :param _datetime start_epoch: Start date/time of files to find and download
+    :param _datetime end_epoch: End date/time of files to find and download
+    :param str file_ext: Extension of files to download (e.g. SP3, CLK, ERP, etc)
+    :param int limit: Variable to limit the number of files downloaded, defaults to None
+    :param bool long_filename: Search for IGS long filename, if None use start_epoch to determine, defaults to None
```
treefern, line 872: This really should be `True` to enable, `False` to disable.
Unless a third value is needed to differentiate other behaviour, in which case to maintain valid typing, it would need wrapping in an `Optional`, i.e. `long_filename: Optional[bool] = None`

gnssanalysis/gn_download.py
```diff
@@ -846,325 +848,183 @@ def download_multiple_files_from_cddis(files: [str], ftp_folder: str, output_fol
...
+    :param str analysis_center: Which analysis center's files to download (e.g. COD, GFZ, WHU, etc), defaults to "IGS"
+    :param str solution_type: Which solution type to download (e.g. ULT, RAP, FIN), defaults to "ULT"
+    :param str sampling_rate: Sampling rate of file to download, defaults to "15M"
+    :param str project_type: Project type of file to download (e.g. ), defaults to "OPS"
+    :param _datetime.timedelta timespan: Timespan of the file/s to download, defaults to _datetime.timedelta(days=2)
+    :param str if_file_present: What to do if file already present: "replace", "dont_replace", defaults to "prompt_user"
+    :raises FileNotFoundError: Raise error if the specified file cannot be found on CDDIS
     """
+    # DZ: Download the correct IGS FIN ERP files
```
EugeneDu-GA, line 881: Can remove `DZ: ` as Des's doc is generic enough?

gnssanalysis/gn_download.py
ronaldmaj: Put in `Optional[bool]`

gnssanalysis/gn_download.py
ronaldmaj: Put in base URLs at top of file

gnssanalysis/gn_download.py
ronaldmaj: Fixed

gnssanalysis/gn_download.py
EugeneDu-GA: And other ftp urls as well? So we can use other ftp/https sites in case the current one is down.

gnssanalysis/gn_download.py
ronaldmaj: Fixed

gnssanalysis/gn_download.py
ronaldmaj: Removed `DZ:`

gnssanalysis/gn_utils.py
```diff
@@ -52,6 +52,31 @@ def get_filetype(path):
     return suffix
 
 
+def configure_logging(verbose: bool) -> None:
+    """Set up the logger object to use for encoding logging strings
+
+    :param bool verbose: Flag to increase level of detail to print to screen - True: DEBUG, False: INFO
+    """
+    if verbose:
```
EugeneDu-GA, line 60: It's confusing (and conflicting) in `clkq` `verbose` being `_logging.INFO` and `else: _logging.disable()`, and so may be somewhere else. I would be better to unify them across the package.

ronaldmaj: Just pushed a commit that uses the `configure_logging` function for all

EugeneDu-GA (approved):

gnssanalysis/gn_download.py
```diff
@@ -459,12 +427,12 @@ def attempt_ftps_download(
...
     """Attempt download of file (filename) given the ftps client object (ftps) to chosen location (download_dir)
 
-    :param _Path download_dir: Path obj to download directory
+    :param _Path download_dir: Where to download files (local directory)
     :param _ftplib.FTP_TLS ftps: FTP_TLS client pointed at download source
     :param str filename: Filename to assign for the downloaded file
     :param str type_of_file: How to label the file for STDOUT messages, defaults to None
     :param str if_file_present: What to do if file already present: "replace", "dont_replace", defaults to "prompt_user"
-    :return _Path: Path obj to the downloaded file
+    :return _Path: Return the pathlib.Path of the downloaded file
```
treefern, line 435:
```suggestion
    :return _Path: The pathlib.Path of the downloaded file
```

gnssanalysis/gn_download.py
```diff
@@ -703,7 +671,7 @@ def download_file_from_cddis(
     :param str if_file_present: What to do if file already present: "replace", "dont_replace", defaults to "prompt_user"
     :param str note_filetype: How to label the file for STDOUT messages, defaults to None
     :raises e: Raise any error that is run into by ftplib
-    :return _Path: Return the local Path of the file downloaded
+    :return _Path: Return the pathlib.Path of the downloaded file
```
treefern, line 674:
```suggestion
    :return _Path: The pathlib.Path of the downloaded file
```
Very minor phrasing detail, it just sounds less redundant dropping the second 'return'

ronaldmaj: No worries, removed the `Return` here and the other 3 places where this appears

gnssanalysis/gn_download.py
```diff
@@ -865,26 +833,30 @@ def download_product_from_cddis(
...
 
-def download_atx(download_dir: _Path, long_filename: bool = False, if_file_present: str = "prompt_user") -> None:
-    """
-    Download the ATX file necessary for running the PEA provided the download directory (download_dir)
-    """
+def download_atx(download_dir: _Path, long_filename: bool = False, if_file_present: str = "prompt_user") -> _Path:
+    """Download the ATX file necessary for running the PEA provided the download directory (download_dir)
 
+    :param _Path download_dir: Where to download files (local directory)
+    :param bool long_filename: Download ATX file relevant after 2022-11-27? True: igs20, False:igs14, defaults to False
```
treefern, line 840:
```suggestion
    :param bool long_filename: Indicates whether to download the newer ATX file (igs20), typically used with long format filenames (post 2022-11-27), or use the older igs14. Defaults to False (igs14).
```

I'm not sure if I've misinterpreted what this was saying, but suffice it to say the meaning wasn't clear. 😅

gnssanalysis/gn_download.py
```diff
@@ -865,26 +833,30 @@ def download_product_from_cddis(
...
+    :param str if_file_present: What to do if file already present: "replace", "dont_replace", defaults to "prompt_user"
+    :return _Path: Return the pathlib.Path of the downloaded file
```
treefern: As above

gnssanalysis/gn_utils.py
```diff
@@ -64,10 +64,11 @@ def configure_logging(verbose: bool, output_logger: bool = False) -> _Union[_log
     else:
         logging_level = _logging.INFO
     _logging.basicConfig(format="%(asctime)s [%(funcName)s] %(levelname)s: %(message)s")
+    _logging.getLogger().setLevel(logging_level)
     if output_logger:
-        return _logging.getLogger().setLevel(logging_level)
+        return _logging.getLogger()
```
treefern, line 69: Good catch. That would've returned `None` whether asked for a logger or not 😆

gnssanalysis/gn_download.py
ronaldmaj: I've changed the name of the argument to make it clearer. We will need to make a subsequent change in the ginan repo for `auto_download_PPP`

tests/test_download.py
```diff
@@ -0,0 +1,89 @@
...
+        self.assertEqual(downloaded_file, None)
+
+    def test_download_yaw_files(self):
+
+        # Test download of yaw files
+        downloaded_files = ga_download.download_yaw_files(download_dir=Path(self.test_dir))
+
+        # Verify
+        self.assertEqual(len(downloaded_files), 3)
+        self.assertEqual(downloaded_files[0].name, "bds_yaw_modes.snx")
```
treefern, line 81: It doesn't really matter either way, but we _probably_ shouldn't care about order here.

We could alternatively check the length of the array, then create a set from it, and check that exactly those three things are present in there.

tests/test_download.py
```diff
@@ -0,0 +1,89 @@
...
+        ga_download.download_product_from_cddis(
+            download_dir=Path(self.test_dir),
+            start_epoch=datetime(2024, 8, 10),
+            end_epoch=datetime(2024, 8, 12),
+            file_ext="SP3",
+        )
+
+        # Verify
+        self.assertEqual(type(next(Path(self.test_dir).glob("*.SP3"))), PosixPath)
+        self.assertEqual(next(Path(self.test_dir).glob("*.SP3")).name, "IGS0OPSULT_20242230000_02D_15M_ORB.SP3")
```
treefern, line 38: If we're primarily interested in it determining the **correct** file to download, I'd suggest shifting the focus to the functions that construct the name, rather than the actual download function (that might require some refactoring if they currently call the download itself).

Alternatively if we do download a specific file, I guess we could also hash it (probably doesn't add much benefit to the test though).

ronaldmaj: For this one, yeah we can definitely have a check on the file size for example, but the other files (ATX, metadata, yaw files) will change from time to time, so not sure what else we can check =

tests/test_download.py
ronaldmaj: Agreed, never know what could happen. I've re-written this to check inside the list rather than assume a specific order

tests/test_download.py
seballgeyer, line 1: In principle, the unittest should be able to run offline.

ronaldmaj: I might just delete `test_download.py` for now, and we can come back to this once we refactor more of the code - we might be able to re-work the download functions to make them easier to unit test

gnssanalysis/gn_download.py
```diff
@@ -846,325 +720,190 @@ def download_multiple_files_from_cddis(files: [str], ftp_folder: str, output_fol
...
+    :param _Path download_dir: Where to download files (local directory)
+    :param str reference_frame: Coordinate reference frame file to download, defaults to "IGS20"
+    :param str if_file_present: What to do if file already present: "replace", "dont_replace", defaults to "prompt_user"
+    :return _Path: The pathlib.Path of the downloaded file
+    """
+    match reference_frame:
+        case "IGS20":
+            atx_filename = "igs20.atx"
+        case "IGb14":
+            atx_filename = "igs14.atx"
```
treefern, line 849: This should perhaps raise a `ValueError` as a default, indicating that the frame name provided was not accepted.

gnssanalysis/gn_utils.py
```diff
@@ -52,6 +52,54 @@ def get_filetype(path):
     return suffix
 
 
+def configure_logging(verbose: bool, output_logger: bool = False) -> _Union[_logging.Logger, None]:
+    """Configure the logger object with the level of verbosity requested and output if desired
+
+    :param bool verbose: Verbosity of logger object to use for encoding logging strings, True: DEBUG, False: INFO
+    :param bool output_logger: Flag to indicate whether to output the Logger object, defaults to False
+    :return _Union[_logging.Logger, None]: Return the logger object or None (based on output_logger)
```
treefern, line 60: These could be switched to Python3.10+ type unions, i.e. `_logging.Logger | None`
```suggestion
    :return _logging.Logger | None: Return the logger object or None (based on output_logger)
```

gnssanalysis/gn_utils.py
```diff
@@ -52,6 +52,54 @@ def get_filetype(path):
     return suffix
 
 
+def configure_logging(verbose: bool, output_logger: bool = False) -> _Union[_logging.Logger, None]:
```
treefern, line 55:
```suggestion
def configure_logging(verbose: bool, output_logger: bool = False) -> _logging.Logger | None:
```

gnssanalysis/gn_download.py
```diff
@@ -403,8 +376,8 @@ def check_whether_to_download(
     already present and what action to take if it is (if_file_present)
 
     :param str filename: Filename of the downloaded file
-    :param _Path download_dir: Path obj to download directory
-    :param str if_file_present: How to handle files that are already present ["replace","dont_replace","prompt_user"], defaults to "prompt_user"
+    :param _Path download_dir: Where to download files (local directory)
+    :param str if_file_present: What to do if file already present: "replace", "dont_replace", defaults to "prompt_user"
     :return _Union[_Path, None]: Path obj to the downloaded file if file should be downloaded, otherwise returns None
```
treefern, line 381: Similarly here

gnssanalysis/gn_download.py
ronaldmaj: Done

gnssanalysis/gn_utils.py
ronaldmaj: Replaced this and other cases where Union was used

gnssanalysis/gn_utils.py
ronaldmaj: Done

tests/test_download.py
EugeneDu-GA: > For this one, yeah we can definitely have a check on the file size for example, but the other files (ATX, metadata, yaw files) will change from time to time, so not sure what else we can check =

Maybe checksum? Or size + last modified time?

treefern (dismissed):

gnssanalysis/gn_download.py
ronaldmaj: Done

treefern (approved):

PR #47, opened 2024-08-23
Transfer Ginan-Ops functions

EugeneDu-GA:
Transfer SP3 and CLK analysis functions from `ssrMonitoring` to `gnssanalysis`
Transfer `generate_sampling_rate()` from `auto_download_PPP` to `gnssanalysis`

### DIFF
--- a/gnssanalysis/gn_diffaux.py
+++ b/gnssanalysis/gn_diffaux.py
@@ -673,6 +673,25 @@ def rac_df_to_rms_df(rac_df):
     return rms_df
 
 
+def format_index(
+    diff_df: _pd.DataFrame,
+) -> None:
+    """
+    Convert the epoch indices of a SP3 or CLK difference DataFrame from J2000 seconds to more readable
+    python datetimes and rename the indices properly.
+
+    :param _pd.DataFrame diff_df: The Pandas DataFrame containing SP3 or CLK differences
+    :return None
+    """
+    # Convert the epoch indices from J2000 seconds to python datetimes
+    diff_df.index = _pd.MultiIndex.from_tuples(
+        ((idx[0] + _gn_const.J2000_ORIGIN, idx[1]) for idx in diff_df.index.values)
+    )
+
+    # Rename the indices
+    diff_df.index = diff_df.index.set_names(["Epoch", "Satellite"])
+
+
 def sp3_diff(
     base_sp3_file: _Path,
     test_sp3_file: _Path,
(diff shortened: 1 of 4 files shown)


gnssanalysis/gn_diffaux.py
```diff
@@ -669,3 +671,155 @@ def rac_df_to_rms_df(rac_df):
...
+    # Convert the epoch indices from J2000 seconds to python datetimes
+    diff_df.index = _pd.MultiIndex.from_tuples(
+        ((idx[0] + _gn_const.J2000_ORIGIN, idx[1]) for idx in diff_df.index.values)
+    )
+
+    # Rename the indices
+    diff_df.index = diff_df.index.set_names(["Epoch", "Satellite"])
+
+
+def sp3_diff(
```
ronaldmaj, line 695: To minimise confusion between this and `diffsp3` perhaps we can call this something else?
`sp3_difference` I think would be enough, but you could go even longer: `sp3_file_difference`

gnssanalysis/gn_diffaux.py
```diff
@@ -669,3 +671,155 @@ def rac_df_to_rms_df(rac_df):
...
+    base_sp3_file: _Path,
+    test_sp3_file: _Path,
+) -> _pd.DataFrame:
+    """
+    Compare two SP3 files to calculate orbit and clock differences. The orbit differences will be represented
+    in both X/Y/Z ECEF frame and R/A/C orbit frame, and the clock differences will NOT be normalised.
```
ronaldmaj, line 701: This also assumes the SP3 file has clock information in it
Does it would when the SP3 file does not have clock info? Or if one does and the other doesnt?

gnssanalysis/gn_diffaux.py
```diff
@@ -669,3 +671,155 @@ def rac_df_to_rms_df(rac_df):
...
+    diff_sp3_df["|Clock|"] = diff_clk_df.abs()
+
+    # Change the epoch indices from J2000 seconds to more readable python datetimes
+    # and rename the indices properly
+    format_index(diff_sp3_df)
+
+    return diff_sp3_df
+
+
+def clk_diff(
```
ronaldmaj, line 745: Again, could look at extending the name of this function

gnssanalysis/gn_diffaux.py
```diff
@@ -669,3 +671,155 @@ def rac_df_to_rms_df(rac_df):
...
+    diff_clk_df["|Clock|"] = diff_clk_df.abs()
+
+    # Change the epoch indices from J2000 seconds to more readable python datetimes
+    # and rename the indices properly
+    format_index(diff_clk_df)
+
+    return diff_clk_df
+
+
+def diff_stats(
```
ronaldmaj, line 777: Something longer could be good here too: `difference_statistics` or something

ronaldmaj (changes requested):
Sorry, I thought I had commented on this previously, but I think this PR is looking good. The one thing I think it needs is a few unit tests on the new functions coming in

gnssanalysis/gn_diffaux.py
EugeneDu-GA: SP3 file always has clock information in it even if it is invalid (e.g. 999999.999999).

ronaldmaj (approved):
Talked about this in-person. Agreed to push through unit-tests in next PRs. My only other comment has been addressed

PR #48, opened 2024-08-23
NPI-3458 implement solution types utility classes

treefern:
This PR adds utility classes to more formally capture the definitions of solution types defined by IGS. This avoids identifying solution types by short strings throughout the codebase, and allows additional details of each solution type to potentially be captured here as well.

### DIFF
--- a/gnssanalysis/enum_meta_properties.py
+++ b/gnssanalysis/enum_meta_properties.py
@@ -1,0 +1,16 @@
+class EnumMetaProperties(type):
+    """
+    This metaclass:
+     - intercepts attempts to set *class* attributes, and rejects them.
+       - NOTE: In the class or abstract class using this, you should also define an __init__() which raises
+         an exception, to prevent instantiation.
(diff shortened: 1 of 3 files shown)


ronaldmaj (commented):
This looks good to me, adds new code but doesn't alter anything existing so won't cause issues. Plus this will help heaps in our Ops space in making things clearer. 

The one thing that could be added are unit tests on these methods, but if this needs to be in the repo to help with testing on Ops, happy for this to go in sooner

ronaldmaj (approved):
Looks good to me, thanks for putting this together! I tested the ability to change the name string after instantiating it and I'm getting the `AttributeError`, so all seems to be working :+1:

treefern:
Updated to more robustly work as static classes, preventing modification of class variables. Some other tidying. Addition of unit tests.
This should be ready to go in now.

ronaldmaj (approved):
Awesome, thanks for putting in the unit tests as well, this will make this a lot more robust to chances in future. Happy for this to go in :+1:

PR #52, opened 2024-09-03
fix: QnD remove time on comparison (calendar date only)

seballgeyer:
Quick and Dirty fix, for the Nanu message selection. 
Now the compareson is make only on the date and not on the date/time. 

Maybe for future improvement, have an option to select if the selection is on date only or datettime. can be good?


ronaldmaj (approved):
Looks good, I've tested this from my side and looks to be working well.

PR #53, opened 2024-09-04
Restoring backward compatibilities

seballgeyer:
Changes in the code to make the package compatible for python 3.9 again. 
Python 3.9 is the default install on Apple MacOS, it's end of life is in October 2025. 

Main changes are:
- moved from match/case statement to dictionary (1 occurence)
- moved from the typehint `|` operator to `Union` 

a few other fixes/simplification the time I was doing it.

### DIFF
--- a/.github/workflows/python-cicd-units.yml
+++ b/.github/workflows/python-cicd-units.yml
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [3.9, 3.10, 3.11, 3.12]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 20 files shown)


ronaldmaj (changes requested):
These changes look great. Also really like that we are now testing these unit tests across various Python versions - this is a really nice addition. 

I've tested this locally on some functions, seem to work well.

The one thing that is coming along for the ride here is a number of `black` changes that I don't think should be coming through.
Namely, compression of lines. We've used line-width = 120 characters for other pushes, so some of the changes in here I don't think should come through if we continue to use that convention

ronaldmaj (approved):
Thanks for going through with `black` :+1: 
I can see it's picked up quite a bit :smile: 
Happy for this to go in

PR #54, opened 2024-09-10
NPI-3501 better empty input file checks

treefern:
These changes should help to avoid misleading downstream errors caused by empty files, which are currently not detected at the file reading stage.

- `path2bytes()` now raises exceptions on all errors rather than returning None. It also raises an `EOFError` if the input data is empty. It has also been restructured slightly for clarity and type checking.
- `read_sp3()` now accepts a `Path` or `bytes` input in addition to a path as `str`. It now writes an empty string into the 'path' attribute if the input was passed as bytes.

### DIFF
--- a/gnssanalysis/gn_io/common.py
+++ b/gnssanalysis/gn_io/common.py
@@ -15,32 +15,43 @@ from typing import Union as _Union
 MB = 1024 * 1024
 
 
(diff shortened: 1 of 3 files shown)


ronaldmaj (approved):
Awesome, we discussed this a bit in person, and it's looking great :+1: 
This will definitely help in debugging issues in Ops a lot quicker

PR #56, opened 2024-09-19
Add download_iau2000_file()

EugeneDu-GA:
Add download_iau2000_file() and other improvements

### DIFF
--- a/gnssanalysis/gn_download.py
+++ b/gnssanalysis/gn_download.py
@@ -894,6 +894,55 @@ def download_product_from_cddis(
                 remain = end_epoch - reference_start
 
 
+def download_iau2000_file(download_dir: _Path, start_epoch: _datetime, if_file_present: str = "prompt_user"):
+    """
+    Download relevant IAU2000 file from CDDIS or IERS based on start_epoch of data
+    """
+    ensure_folders([download_dir])
+    # Download most recent daily IAU2000 file if running for a session within the past week (data is within 3 months)
+    if _datetime.datetime.now() - start_epoch < _datetime.timedelta(weeks=1):
+        url_dir = "daily/"
+        iau2000_filename = "finals2000A.daily"
+        download_filename = "finals.daily.iau2000.txt"
+        logging.info("Attempting Download of finals2000A.daily file")
+    # Otherwise download the IAU2000 file dating back to 1992
+    else:
+        url_dir = "standard/"
+        iau2000_filename = "finals2000A.data"
+        download_filename = "finals.data.iau2000.txt"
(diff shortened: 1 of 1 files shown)


gnssanalysis/gn_download.py
```diff
@@ -830,12 +831,15 @@ def download_product_from_cddis(
     logging.debug(
         f"Generated filename: {product_filename}, with GPS Date: {gps_date.gpswkD} and reference: {reference_start}"
     )
+
+    ensure_folders([download_dir])
+    download_filepaths = []
```
ronaldmaj, line 836: This is not returned at the end (if that's what we want to change the behaviour to)

gnssanalysis/gn_download.py
```diff
@@ -877,17 +881,68 @@ def download_product_from_cddis(
...
+                            note_filetype=file_ext,
+                        )
                     )
                 count += 1
                 remain = end_epoch - reference_start
 
 
+def download_iau2000_file(download_dir: _Path, start_epoch: _datetime, if_file_present: str = "prompt_user"):
+    """
+    Download relevant IAU2000 file from CDDIS or IERS based on start_epoch of data
```
ronaldmaj, line 899: docstring will need to be expanded out here

gnssanalysis/gn_download.py
```diff
@@ -503,27 +503,27 @@ def attempt_ftps_download(
 
 
 def attempt_url_download(
-    download_dir: _Path, url: str, filename: str = None, type_of_file: str = None, if_file_present: str = "prompt_user"
+    download_dir: _Path, url: str, download_filename: str = None, type_of_file: str = None, if_file_present: str = "prompt_user"
```
ronaldmaj, line 506: If we are changing `download_filename` here, we'll need to make sure all functions that call this, use the new argument name

ronaldmaj (changes requested):
This is looking good. The changes to the `attempt_download` function look ok, but we'll need to make sure that all functions that rely on it are ok

gnssanalysis/gn_download.py
EugeneDu-GA: Changed back to `filename` as it might affect other external scripts including `auto_download_PPP.py`

ronaldmaj (changes requested):
Looking really good, thanks for addressing my previous comments. Just a formatting issue I found in that the line length seems not to be set at 120? After that I think this can go in

ronaldmaj (approved):
Cool, thanks for fixing up the formatting. I think this is good to go in now

PR #57, opened 2024-09-19
SP3 clock normalisation & Helmert transformation

EugeneDu-GA:
Enable clock normalisation and Helmert transformation with sp3_difference


ronaldmaj (approved):
This looks good to me. I've done a check for other functions that call `gn_io.clk.sv_clocks` but I don't see anything else.
There is already a unit test for `compare_clk()` so that's covered, but I cannot say anything about `sp3_difference()` or `clk_difference()`. If it's working as expected for your purposes then I think that's fine.
I would look at putting through unit tests for these two functions (doesn't have to be in this PR)

gnssanalysis/gn_io/sp3.py
```diff
@@ -809,17 +809,36 @@ def sp3merge(
...
+
+    :param DataFrame a: The sp3_a DataFrame.
+    :param DataFrame b : The sp3_b DataFrame.
+    :param bool epochwise: Epochwise HLM transformation.
+
+    :return tuple[pandas.DataFrame, list]: A tuple containing the updated sp3_b DataFrame and the HLM array with applied computed parameters and residuals.
+    """
+    if epochwise:
+        for epoch in b.index.get_level_values("J2000").unique():
+            b.loc[epoch].iloc[:, :3], hlm = hlm_trans(a.loc[epoch].EST[["X", "Y", "Z"]].values, b.loc[epoch].EST[["X", "Y", "Z"]].values)  # Eugene: hlm will be overwritten in the loop
```
seballgeyer, line 838: I suppose you want hlm to be define first (before the if statement)
```python
hlm=[]
```
and in the for loop have:
```python
b.loc[epoch].iloc[:, :3], hlm_epoch = .... 
```
followed by 
```python
hlm.append(hlm_epoch)
```

seballgeyer (changes requested):

ronaldmaj (approved):
As I said before, I think this is in a good enough state to go in, but I would wait on @seballgeyer to give his final approval before merging in

The one thing I would like to see is unit tests on functions that were touched here, but perhaps that could be done in a more wide-ranging PR that creates unit tests for a whole range of functions

seballgeyer (approved):

PR #59, opened 2024-10-16
NPI-3547 Attempt to address SP3 flag rendering related crash

treefern:
This change:
- Adds checks for SP3 flag values (during SP3 writing).
- Makes datatypes explicit while concatenating flags into a single column for rendering reasons. Hopefully a recent crash related to that will be addressed by this change (see corresponding ticket number for details).


ronaldmaj (approved):
Looks good to me! A nice check on the flags and ensuring that the file information should follow conventions. Talked with @treefern about his testing process. Happy for this to go in 👍

PR #60, opened 2024-12-12
NPI-3653 - Update code to handle new V2 IGS logs

ronaldmaj:
A quick PR to update the code to handle new metadata fields in IGS log files.

This will still be able to read and process the old IGS log files as well (backwards compatible)

### DIFF
--- a/gnssanalysis/gn_io/igslog.py
+++ b/gnssanalysis/gn_io/igslog.py
@@ -159,7 +159,7 @@ def parse_igs_log(filename_array: _np.ndarray) -> _np.ndarray:
     :param _np.ndarray filename_array: Metadata on input log file. Expects ndarray of the form [CODE DATE PATH]
     :return _np.ndarray: Returns array with data from the IGS log file parsed
     """
-    file_code, __, file_path = filename_array
+    file_code, _, file_path = filename_array
 
     with open(file_path, "rb") as file:
         data = file.read()
--- a/gnssanalysis/gn_io/igslog.py
+++ b/gnssanalysis/gn_io/igslog.py
@@ -294,13 +294,13 @@ def translate_series(series: _pd.Series, translation: dict) -> _pd.Series:
 
 def gather_metadata(
     logs_glob_path: str = "/data/station_logs/station_logs_IGS/*/*.log", rnx_glob_path: str = None, num_threads: int = 1
-) -> _List[_pd.DataFrame, _pd.DataFrame, _pd.DataFrame]:
+) -> _List[_pd.DataFrame]:
     """Parses log files found with glob expressions into pd.DataFrames
 
     :param str logs_glob_path: A glob expression for log files, defaults to "/data/station_logs_IGS/*/*.log"
     :param str rnx_glob_path: A glob expression for rnx files, e.g. /data/pea/exs/data/*.rnx, defaults to None
     :param int num_threads: Number of threads to run, defaults to 1
-    :return _List[_pd.DataFrame, _pd.DataFrame, _pd.DataFrame]: List of DataFrames with [ID, Receiver, Antenna] data
+    :return _List[_pd.DataFrame]: List of DataFrames with [ID, Receiver, Antenna] data
     """
     parsed_filenames = find_recent_logs(logs_glob_path=logs_glob_path, rnx_glob_path=rnx_glob_path).values
 
--- a/gnssanalysis/gn_io/igslog.py
+++ b/gnssanalysis/gn_io/igslog.py
@@ -1,5 +1,6 @@
 """IGS log files parser"""
 
+import logging
 import glob as _glob
 import re as _re
 from multiprocessing import Pool as _Pool
@@ -12,6 +13,8 @@ from .. import gn_frame as _gn_frame
 from .. import gn_io as _gn_io
 from .. import gn_transform as _gn_transform
 from tqdm import tqdm
+
+logger = logging.getLogger(__name__)
 
 _REGEX_ID = _re.compile(
     rb"""
@@ -126,6 +129,22 @@ def find_recent_logs(logs_glob_path: str, rnx_glob_path=None) -> _pd.DataFrame:
     return recent_logs_df
 
 
+def determine_log_version(data: bytes):
+    REGEX_VERSION_1 = _re.compile(rb"""(site log\))""")
+    REGEX_VERSION_2 = _re.compile(rb"""(site log v2)""")
+
+    # check for version 1:
+    result_v1 = REGEX_VERSION_1.search(data)
+    result_v2 = REGEX_VERSION_2.search(data)
+
+    if result_v1 is not None:
+        return "v1.0"
+    elif result_v2 is not None:
+        return "v2.0"
+    else:
+        return "Unknown"
+
+
 def parse_igs_log(filename_array: _np.ndarray) -> _np.ndarray:
     """Parses igs log and outputs ndarray with parsed data
     Expects ndarray of the form [CODE DATE PATH]"""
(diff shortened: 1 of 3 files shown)


gnssanalysis/gn_io/igslog.py
```diff
@@ -109,21 +134,25 @@ def parse_igs_log(filename_array: _np.ndarray) -> _np.ndarray:
...
         data = file.read()
 
-    blk_id = _REGEX_ID.search(data)
+    blk_id = _REGEX_ID_V2.search(data)
     if blk_id is None:
-        tqdm.write(f"ID rejected from {file_path}")
-        return _np.array([]).reshape(0, 12)
+        blk_id = _REGEX_ID.search(data)
+        if blk_id is None:
+            tqdm.write(f"ID rejected from {file_path}")
```
seballgeyer, line 141: There is just 2 files in the library that uses tqdm (top.py too), shall we think about removing it?

gnssanalysis/gn_io/igslog.py
```diff
@@ -109,21 +134,25 @@ def parse_igs_log(filename_array: _np.ndarray) -> _np.ndarray:
     with open(file_path, "rb") as file:
         data = file.read()
 
-    blk_id = _REGEX_ID.search(data)
+    blk_id = _REGEX_ID_V2.search(data)
     if blk_id is None:
```
seballgeyer, line 138: Maybe a variable `version_2` can be used to have so we don't need to do the check for the next modification

gnssanalysis/gn_io/igslog.py
```diff
@@ -109,21 +134,25 @@ def parse_igs_log(filename_array: _np.ndarray) -> _np.ndarray:
...
 
     blk_id = [blk_id[1].decode().upper(), blk_id[2].decode().upper()]  # no .groups() thus 1 and 2
     code = blk_id[0]
     if code != file_code:
         tqdm.write(f"{code}!={file_code} at {file_path}")
         return _np.array([]).reshape(0, 12)
 
-    blk_loc = _REGEX_LOC.search(data)
+    blk_loc = _REGEX_LOC_V2.search(data)
     if blk_loc is None:
```
seballgeyer, line 151: if we have a variable version2 we can just do 
```
if version2:
  do something
else:
  do something else

File: gnssanalysis/gn_io/igslog.py [hunk 4]
```diff
@@ -109,21 +134,25 @@ def parse_igs_log(filename_array: _np.ndarray) -> _np.ndarray:
     with open(file_path, "rb") as file:
         data = file.read()
 
-    blk_id = _REGEX_ID.search(data)
+    blk_id = _REGEX_ID_V2.search(data)
```
seballgeyer, line 137: an unit test can be good so we can have a quick check if it works or not?

gnssanalysis/gn_io/igslog.py
```diff
@@ -109,21 +134,25 @@ def parse_igs_log(filename_array: _np.ndarray) -> _np.ndarray:
     with open(file_path, "rb") as file:
```
seballgeyer, line 134: [suggestion]
```suggestion
    with open(file_path, "rb") as file:
```
The function itself is doing as a few goals. It will be good to separate the concerns so it is doing specific goals. 
for example
```
def parse_igs_log:
   call to read igs file
   cal to find V1 or V2
   them call specific funtion to read each parts.

File: gnssanalysis/gn_io/igslog.py [hunk 6]
```diff
@@ -109,21 +134,25 @@ def parse_igs_log(filename_array: _np.ndarray) -> _np.ndarray:
...
         data = file.read()
```
seballgeyer, line 135: Also as you like to comment on everyone else PR. 
the time you change this file, could you review and update the documentation `on-line sphinx style` with input/output descriptions ;)

seballgeyer (changes requested):

gnssanalysis/gn_io/igslog.py
```diff
@@ -101,6 +129,22 @@ def find_recent_logs(logs_glob_path: str, rnx_glob_path=None) -> _pd.DataFrame:
     return recent_logs_df
 
 
+def determine_log_version(data: bytes):
+    REGEX_VERSION_1 = _re.compile(rb"""(site log\))""")
```
seballgeyer, line 133: REGEX_VERSION_1 should be _REGEX_VERSION_1 to follow the existing coding standard

gnssanalysis/gn_io/igslog.py
```diff
@@ -101,6 +129,22 @@ def find_recent_logs(logs_glob_path: str, rnx_glob_path=None) -> _pd.DataFrame:
...
+    REGEX_VERSION_2 = _re.compile(rb"""(site log v2)""")
+
+    # check for version 1:
+    result_v1 = REGEX_VERSION_1.search(data)
```
seballgeyer, line 137: I suppose if version 1 is successful we don't need to check if version 2 is good?

gnssanalysis/gn_io/igslog.py
```diff
@@ -101,6 +129,22 @@ def find_recent_logs(logs_glob_path: str, rnx_glob_path=None) -> _pd.DataFrame:
...
+    result_v2 = REGEX_VERSION_2.search(data)
+
+    if result_v1 is not None:
```
seballgeyer, line 140: `if result_v1 is not None` should have the same effect then `if result_v1`

gnssanalysis/gn_io/igslog.py
```diff
@@ -101,6 +129,22 @@ def find_recent_logs(logs_glob_path: str, rnx_glob_path=None) -> _pd.DataFrame:
     return recent_logs_df
 
 
+def determine_log_version(data: bytes):
```
seballgeyer, line 132: return type missing.

seballgeyer (changes requested):

gnssanalysis/gn_io/igslog.py
ronaldmaj: Done

gnssanalysis/gn_io/igslog.py
ronaldmaj: Left this as is, introduced an if/else above to assign `_REGEX_ID` ...

gnssanalysis/gn_io/igslog.py
ronaldmaj: Achieved with new function search for version number, and then assigning existing variables based on the regex pattern to use

gnssanalysis/gn_io/igslog.py
ronaldmaj: Fixed

gnssanalysis/gn_io/igslog.py
ronaldmaj: Fixed.
And yep, there is a closed bracket in version 1 - I guess they weren't expecting it to change

gnssanalysis/gn_io/igslog.py
ronaldmaj: Yep true. This has been updated so that a check of both is no longer necessary

gnssanalysis/gn_io/igslog.py
ronaldmaj: Changed to `if result_v1`

gnssanalysis/gn_io/igslog.py
```diff
@@ -101,6 +129,22 @@ def find_recent_logs(logs_glob_path: str, rnx_glob_path=None) -> _pd.DataFrame:
...
+    # check for version 1:
+    result_v1 = REGEX_VERSION_1.search(data)
+    result_v2 = REGEX_VERSION_2.search(data)
+
+    if result_v1 is not None:
+        return "v1.0"
+    elif result_v2 is not None:
+        return "v2.0"
+    else:
+        return "Unknown"
```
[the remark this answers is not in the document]
ronaldmaj: Added

gnssanalysis/gn_io/igslog.py
ronaldmaj: I've moved the "find version number" functionality out and it is now just called

gnssanalysis/gn_io/igslog.py
ronaldmaj: I've gone through and updated the doc-strings for all functions in this file

gnssanalysis/gn_io/igslog.py
```diff
@@ -109,30 +153,39 @@ def parse_igs_log(filename_array: _np.ndarray) -> _np.ndarray:
     with open(file_path, "rb") as file:
         data = file.read()
 
-    blk_id = _REGEX_ID.search(data)
+    version = determine_log_version(data)
```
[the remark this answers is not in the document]
ronaldmaj: Done and done.

ronaldmaj (commented):
I think I've address the outstanding comments you've left on this. Let me know if I've missed anything

gnssanalysis/gn_io/igslog.py
```diff
@@ -101,38 +138,76 @@ def find_recent_logs(logs_glob_path: str, rnx_glob_path=None) -> _pd.DataFrame:
...
+    """
+    file_code, _, file_path = filename_array
 
     with open(file_path, "rb") as file:
         data = file.read()
 
+    try:
+        version = determine_log_version(data)
+    except LogVersionError as e:
+        logger.warning(f"Error: {e}, skipping parsing the log file")
```
seballgeyer, line 178: in this case, when the exception is raised, the code will still continue even if the version is uknown, 
I am not sure what do we want to do. 
one option is to stop the function (either re- `raise` 
```python
except LogVersionError as e:
    logger.warning(...)
   raise
```
or just an empty return
```python
except .....
     logger.warning(....)
    return
```

seballgeyer (changes requested):

gnssanalysis/gn_io/igslog.py
ronaldmaj: I've gone with returning None

gnssanalysis/gn_io/igslog.py
ronaldmaj: I've now written up unit tests for all the regex functions in `parse_igs_log()` - which includes the two new ones I've introduced.
I was considering whether to just do a unit test for the entire `parse_igs_log()` function, but I've decided to break it up.
As always, happy to try other suggestions

tests/test_igslog.py
```diff
@@ -0,0 +1,73 @@
...
+
+from gnssanalysis.gn_io import igslog
+from test_datasets.sitelog_test_data import abmf_site_log_v1 as v1_data, abmf_site_log_v2 as v2_data
+
+
+class Testregex(unittest.TestCase):
+    def test_determine_log_version(self):
+        # Ensure version 1 and 2 strings are produced as expected
+        self.assertEqual(igslog.determine_log_version(v1_data), "v1.0")
+        self.assertEqual(igslog.determine_log_version(v2_data), "v2.0")
```
seballgeyer, line 13: Missing checking if it raise an exception if no igs log is provided
```python
self.assertRaises
```

seballgeyer (changes requested):

seballgeyer (dismissed):

seballgeyer (approved):

PR #61, opened 2024-12-18
NPI-3655 Bring in remove_offline_sats(), add option in SP3 diff processing to avoid crashing cubic interpolator for velocity

treefern:
This PR brings in `remove_offline_sats()` from the other codebase, and adds options to `diff_sp3_rac()` to remove offline / partial nodata sats before running an SP3 diff. This is important because `NaN` / non-finite position data crashes the cubic interpolator used for velocity calculations.

A warning is now logged when using the cubic interpolator but **not** also enabling the option to remove offline sats.


gnssanalysis/gn_io/sp3.py
```diff
@@ -191,6 +191,23 @@ def sp3_clock_nodata_to_nan(sp3_df: _pd.DataFrame) -> None:
     sp3_df.loc[nan_mask, ("EST", "CLK")] = _np.nan
 
 
+def remove_offline_sats(sp3_df: _pd.DataFrame):
```
seballgeyer, line 194: Can we have an unit test on this one to make sure it works as expected?
Needs maybe 2 epochs, 2 satellite one with `0`.

seballgeyer (changes requested):

treefern:
Further testing revealed this wasn't working as intended, due to only handling the raw representation of POS nodata, not handling `NaN`s. Updated.

seballgeyer (approved):
Looks good now. 
Did some very quick test.

PR #62, opened 2024-12-18
NPI-3655 Improve readability of getVelPoly() function

treefern:
No functional changes, but a bunch of renamed variables to hopefully make it clearer what this function is doing.

### DIFF
--- a/gnssanalysis/gn_io/sp3.py
+++ b/gnssanalysis/gn_io/sp3.py
@@ -477,28 +477,28 @@ def getVelPoly(sp3Df: _pd.DataFrame, deg: int = 35) -> _pd.DataFrame:
 
     """
     est = sp3Df.unstack(1).EST[["X", "Y", "Z"]]
-    x = est.index.get_level_values("J2000").values
-    # x = est.index.values
-    y = est.values
+    time_values = est.index.get_level_values("J2000").values
+    xyz_positions = est.values
 
-    off, scl = mapparm([x.min(), x.max()], [-1, 1])  # map from input scale to [-1,1]
+    # map from input scale to [-1,1]
+    offset, scale_factor = mapparm([time_values.min(), time_values.max()], [-1, 1])
 
-    x_new = off + scl * (x)
-    coeff = _np.polyfit(x=x_new, y=y, deg=deg)
+    normalised_time_values = offset + scale_factor * (time_values)
+    coeff = _np.polyfit(x=normalised_time_values, y=xyz_positions, deg=deg)
 
-    x_prev = off + scl * (x - 1)
-    x_next = off + scl * (x + 1)
+    time_prev = offset + scale_factor * (time_values - 1)
+    time_next = offset + scale_factor * (time_values + 1)
 
-    xx_prev_combined = _np.broadcast_to((x_prev)[None], (deg + 1, x_prev.shape[0]))
-    xx_next_combined = _np.broadcast_to((x_next)[None], (deg + 1, x_prev.shape[0]))
+    time_prev_sqrd_combined = _np.broadcast_to((time_prev)[None], (deg + 1, time_prev.shape[0]))
+    time_next_sqrd_combined = _np.broadcast_to((time_next)[None], (deg + 1, time_prev.shape[0]))
 
-    inputs_prev = xx_prev_combined ** _np.flip(_np.arange(deg + 1))[:, None]
-    inputs_next = xx_next_combined ** _np.flip(_np.arange(deg + 1))[:, None]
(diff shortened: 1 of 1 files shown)


gnssanalysis/gn_io/sp3.py
```diff
@@ -477,28 +477,28 @@ def getVelPoly(sp3Df: _pd.DataFrame, deg: int = 35) -> _pd.DataFrame:
...
     est = sp3Df.unstack(1).EST[["X", "Y", "Z"]]
-    x = est.index.get_level_values("J2000").values
-    # x = est.index.values
-    y = est.values
+    time_values = est.index.get_level_values("J2000").values
+    xyz_positions = est.values
 
-    off, scl = mapparm([x.min(), x.max()], [-1, 1])  # map from input scale to [-1,1]
+    # map from input scale to [-1,1]
+    offset, scale_factor = mapparm([time_values.min(), time_values.max()], [-1, 1])
```
seballgeyer, line 484: not sure we need `factor` scale can be enough

seballgeyer (commented):
It is a good idea to make it clearer. 
However sometimes long variable name can make the code more difficult to read. And here some a quite long.

I did a few suggestions, I didn't went through all changes.  

Otherwise it is OK

gnssanalysis/gn_io/sp3.py
```diff
@@ -477,28 +477,28 @@ def getVelPoly(sp3Df: _pd.DataFrame, deg: int = 35) -> _pd.DataFrame:
 
     """
     est = sp3Df.unstack(1).EST[["X", "Y", "Z"]]
-    x = est.index.get_level_values("J2000").values
-    # x = est.index.values
-    y = est.values
+    time_values = est.index.get_level_values("J2000").values
+    xyz_positions = est.values
```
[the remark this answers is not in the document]
treefern: 👍 `positions` would be fine yeah. My original concern was that the variables being `x` and `y`, could've sounded like the x and y _components_ of positions.

gnssanalysis/gn_io/sp3.py
```diff
@@ -477,28 +477,28 @@ def getVelPoly(sp3Df: _pd.DataFrame, deg: int = 35) -> _pd.DataFrame:
 
     """
     est = sp3Df.unstack(1).EST[["X", "Y", "Z"]]
-    x = est.index.get_level_values("J2000").values
-    # x = est.index.values
-    y = est.values
+    time_values = est.index.get_level_values("J2000").values
```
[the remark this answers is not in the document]
treefern: `times` I guess; I just wanted to make it clearer that it's a series.
Will update 👍

gnssanalysis/gn_io/sp3.py
treefern: Just noticed it was `scale_factor` in some other similar code, which got introduced during a PR review:

https://github.com/GeoscienceAustralia/gnssanalysis/commit/fe81af2bef28727413acccf89b8c43a778d33aea#diff-330d45d4cfeafdbd9dcc218de027fe2763d140a22cf9763c6cf944f9ad0f1543R108

gnssanalysis/gn_io/sp3.py
```diff
@@ -477,28 +477,28 @@ def getVelPoly(sp3Df: _pd.DataFrame, deg: int = 35) -> _pd.DataFrame:
...
 
-    x_new = off + scl * (x)
-    coeff = _np.polyfit(x=x_new, y=y, deg=deg)
+    normalised_times = offset + scale_factor * (times)
+    coeff = _np.polyfit(x=normalised_times, y=positions, deg=deg)
 
-    x_prev = off + scl * (x - 1)
-    x_next = off + scl * (x + 1)
+    time_prev = offset + scale_factor * (times - 1)
+    time_next = offset + scale_factor * (times + 1)
```
treefern, line 490: Should this actually be called `times_next`? Given it appears to be *all* the time values, shifted by one step forward.
@seballgeyer

seballgeyer: More a look into this function, obscurer it looks... 
it seems that the 20 lines functions is doing something that numpy can do by itself in 2 or 3 function calls. 

I would suggest to merge this PR and eventually rewrite this function later.

seballgeyer (approved):
More a look into this function, obscurer it looks...
it seems that the 20 lines functions is doing something that numpy can do by itself in 2 or 3 function calls.

I would suggest to merge this PR and eventually rewrite this function later. The objectives of renaming some variable to improve understanding has been achieved.

PR #64, opened 2024-12-23
NPI-3668 SP3 header stated values vs content checks, and ORB_TYPE update on performing Helmert transformation

treefern:
This PR is intended to add a bit more robustness to our generated SP3 headers.

This adds:
- Checks that the list of SVs the header claims the SP3 data contains, match the values in the DataFrame itself (which will become the content of the SP3 file written out).
- Logic to set the `ORB_TYPE` field of the SP3 header to `HLM`, when a Helmert transformation is performed on the SP3 DataFrame. This is done to keep the header metadata used to write out headers, more consistent with changes made to the data itself.
- Some additional unit test assertions relating to header SV count and SV list length, and merging of SV accuracy codes (*lowest common denominator*, in effect).
- Various clarifying comments and variable name changes


tests/test_sp3.py
```diff
@@ -223,3 +223,17 @@ def test_sp3merge(self):
         self.assertEqual(result.attrs["HEADER"].HEAD.AC, "AIES")
         self.assertEqual(result.attrs["HEADER"].HEAD.COORD_SYS, None)
         self.assertEqual(result.attrs["HEADER"].HEAD.PV_FLAG, "P")
+        self.assertEqual(
```
ronaldmaj, line 226: This is a nice test to see that the feature works as intended for a normal file, but it would be good to write one up to check that it also works for a corrupted file / header (i.e. get the error to flag as expected)

gnssanalysis/gn_io/sp3.py
```diff
@@ -880,10 +929,13 @@ def sp3_hlm_trans(a: _pd.DataFrame, b: _pd.DataFrame) -> tuple[_pd.DataFrame, li
...
      :param DataFrame b : The sp3_b DataFrame.
 
-    :returntuple[pandas.DataFrame, list]: A tuple containing the updated sp3_b DataFrame and the HLM array with applied computed parameters and residuals.
+    :return tuple[pandas.DataFrame, list]: A tuple containing the updated sp3_b DataFrame and the HLM array with
+        applied computed parameters and residuals. The ORB_TYPE header of the output DataFrame is updated to HLM.
     """
     hlm = _gn_transform.get_helmert7(pt1=a.EST[["X", "Y", "Z"]].values, pt2=b.EST[["X", "Y", "Z"]].values)
     b.iloc[:, :3] = _gn_transform.transform7(xyz_in=b.EST[["X", "Y", "Z"]].values, hlm_params=hlm[0])
+
+    b.attrs["HEADER"].head.ORB_TYPE = "HLM"  # Update b's header to reflect Helmert transformation has been applied
```
ronaldmaj, line 938: This is a good addition to ensure we are conveying accurate information

ronaldmaj (changes requested):
This is a great addition and something we've discussed so generally I'm happy with how this has turned out. 
My one suggestion is to include a test for a file / header where the number of satellites is not correct - make sure the error flags as expected. 
I think just a small header should be enough, one where the number in the header is wrong. Then you can pass that bytes object through to `parse_sp3_header`.

treefern:
Added some additional consistency check logic to `read_sp3()`.
In particular new consistency checks on the number of epochs given by the filename / header / content, plus an initial unit test for this (should be extended with more cases and a success example).

ronaldmaj (approved):
Happy for this to go in - unit-test to check an sp3 block with incorrect number of SVs is now in the code

PR #66, opened 2025-01-07
NPI-3676 Exception handling on (unsupported) SP3 velocity data output

treefern:
`gnssanalysis` currently does not correctly support output of SP3 velocity data.

At present, currently calling `gen_sp3_content()` on a `DataFrame` containing velocity data, will result in a non-compliant SP3 file with velocity data in additional columns (not in additional, interlaced rows as the spec calls for).

This PR adds a check for velocity columns. If any velocity columns are found a warning is logged, the columns are deleted, then output continues as normal on just the position data.

Optionally based on a new parameter to `gen_sp3_content()`, a `NotImplementedError` can be raised instead of logging a warning.

### DIFF
--- a/gnssanalysis/gn_io/sp3.py
+++ b/gnssanalysis/gn_io/sp3.py
@@ -642,6 +642,7 @@ def gen_sp3_content(
     sp3_df: _pd.DataFrame,
     sort_outputs: bool = False,
     buf: Union[None, _io.TextIOBase] = None,
+    continue_on_unhandled_velocity_data: bool = True,
 ) -> str:
     """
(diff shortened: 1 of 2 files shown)


treefern:
A small fix for the `sp3_merge()` unit test came along for the ride.
It turns out that pyfakefs actually uses a real temporary directory under the hood. It seems this can lead to data persisting between unit test runs!

ronaldmaj (approved):
This is a nice clean-up PR and sets up a new unit-test for future. Happy for this to go in

ronaldmaj (approved):
All good from what I can see after the merge with `main` - good to go

PR #70, opened 2025-01-13
NPI-3683 SP3 consistency check streamlining

treefern:
This PR adds various restructuring of the SP3 read and validation process, aimed at improving consistency and separating concerns better between stages of the process.

Also:
- Stopped setting header `ORB_TYPE` to `INT` when removing offline sats.
- Added utility functions to derive unique epochs and SVs, to capture the nuances in a single place.
- Added `remove_svs_from_header()` utility function for removing sats from internal header, so we can better maintain header consistency when performing operations like `remove_offline_sats()`
- Improved `filter_by_svs()`, reworking the removal logic to only call `drop()` once, and to update the internal header to reflect the new list of sats.
- Added experimental detection and warnings for old / new versions of SP3 format.

### DIFF
--- a/gnssanalysis/gn_io/sp3.py
+++ b/gnssanalysis/gn_io/sp3.py
@@ -547,7 +547,6 @@ def read_sp3(
     header = content[:header_end]
     content = content[header_end:]
     parsed_header = parse_sp3_header(header)
-    header_sv_count = parsed_header.SV_INFO.count()  # count() of non-NA/null, vs shape: dims of whole structure
     fline_b = header.find(b"%f") + 2  # TODO add to header parser
     fline = header[fline_b : fline_b + 24].strip().split(b"  ")
     base_xyzc = _np.asarray([float(fline[0])] * 3 + [float(fline[1])])  # exponent base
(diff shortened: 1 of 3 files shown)


gnssanalysis/gn_io/sp3.py
```diff
@@ -193,14 +196,41 @@ def sp3_clock_nodata_to_nan(sp3_df: _pd.DataFrame) -> None:
     sp3_df.loc[nan_mask, ("EST", "CLK")] = _np.nan
 
 
+def remove_svs_from_header(sp3_df: _pd.DataFrame, sats_to_remove: set[str]):
```
ronaldmaj, line 199: Missing output type hint

gnssanalysis/gn_io/sp3.py
```diff
@@ -193,14 +196,41 @@ def sp3_clock_nodata_to_nan(sp3_df: _pd.DataFrame) -> None:
...
+    """
+    Utility function to update the internal representation of an SP3 header, when SVs are removed from the SP3
+    DataFrame. This is useful e.g. when removing offline satellites.
+    :param _pd.DataFrame sp3_df: SP3 DataFrame on which to update the header (in place).
+    :param list[str] sats_to_remove: list of SV names to remove from the header.
+    """
```
ronaldmaj, line 205: doc string missing output parameter

ronaldmaj (changes requested):
A nice general clean-up to the sp3 reading / editing functionality, plus relevant unit-tests.
I've run the tests locally, all seems to work well. I don't see any major issues, just one type-hint / docstring to address. Other than that I think it's good to go

gnssanalysis/gn_io/sp3.py
treefern: There is none; this function operates on the `DataFrame` passed in, modifying it in place.

ronaldmaj (approved):
Thanks for putting through my requested changes. Good to go in 👍

PR #71, opened 2025-01-13
NPI-3685 Streamline SP3 incorrect timerange unit test

treefern:
This PR attempts to speed up one of our slower unit tests, which parses a full day worth of SP3 data.

The number of sats in the test file has been dropped from 30 to 1.
Test run time is still slow, but perhaps 1-2 seconds faster than it was.

After a bit of investigation, it appears the main bottleneck in reading SP3 files, is the fact ***each epoch*** is parsed into a temporary `DataFrame`, before being concatenated into the output `DataFrame`. So, without significantly overhauling the relevant code, number of epochs is likely to be the main determiner of speed.

Also added:
- a small and experimental utility class `ContextTimer`, which can be used as a context manager, to time function executions.
- use of pyfakefs `reset()` in various unit tests, to better guard against data from previous test runs impacting tests.


ronaldmaj (approved):
This looks good to me - nothing major to comment on. The addition of the `reset()` method is great to lower chances of issues with our unit tests in future

PR #72, opened 2025-01-13
NPI-3388 Pandas deprecation fixes and improved SP3 rendering capability (moon orbit capability 🌚 )

treefern:
This PR has been a while in the making. 😅 
The motivation for this work, was **fixing deprecation issues which would prevent upgrading to Pandas 3.**

One of these issues required changing the way `gen_sp3_content()` renders values within each epoch.
We previously found and replaced `NaN`s with string nodata values, within the `DataFrame` itself. This represented double-handling and is also not allowable in Pandas 3 as it mixed datatypes within a column (we were putting a string nodata value in a numeric column when we found a `NaN` or `+/-inf`).
 
We how handle formatting of `inf` and `NaN` values at the point of rendering, using `Styler`'s `format()`, `hide()`, and `to_string()` functions. This means less double-handling and should improve performance and code clarity. ✨ 

Switching to `Styler` also had an unexpected bonus 🎉 : `Styler`'s `to_string()` has a `delimiter` parameter, and setting this to an empty string allows us to **disable padding between columns** (which had previously been a bugbear).
This change in turn **allows us to format position and clock values at the correct width**, removing a known limitation in the SP3 rendering code.
Now we can output a file with an orbit as large as that of the moon. 🌚 But we'd probably never need to.


treefern:
Further update made to address misalignment of values with full width enabled. It turns out the nodata strings needed updating.

ronaldmaj (approved):
This is awesome, I've just tested this against the unit-tests and all deprecations warnings are now gone (from what I can see). Very happy for this to go in 👍

PR #74, opened 2025-01-23
NPI-3689 Restructure IAU2000 download functionality

treefern:
Separate IAU2000 file variant selection logic from download functionality and improve selection logic to allow passing a date range rather than a single date.

### DIFF
--- a/gnssanalysis/gn_download.py
+++ b/gnssanalysis/gn_download.py
@@ -931,17 +931,17 @@ def download_product_from_cddis(
     return download_filepaths
 
 
-def download_iau2000_varient(
+def download_iau2000_variant(
     download_dir: _Path,
-    iau2000_file_varient: Literal["standard", "daily"],
+    iau2000_file_variant: Literal["standard", "daily"],
     if_file_present: str = "prompt_user",
 ) -> Union[_Path, None]:
     """
-    Downloads IAU2000 file based on the varient requested ("daily" or "standard" file).
+    Downloads IAU2000 file based on the variant requested ("daily" or "standard" file).
     Added in approximately version 0.0.58
 
     :param _Path download_dir: Where to download file (local directory).
-    :param Literal["standard", "daily"] iau2000_file_varient: name of file varient to download. Specifies whether to
+    :param Literal["standard", "daily"] iau2000_file_variant: name of file variant to download. Specifies whether to
         download the recent "daily" file, or the historical "standard" file.
     :param str if_file_present: What to do if file already present: "replace", "dont_replace", defaults to "prompt_user"
     :raises Exception: On download failures. In some cases the underlying exception will be wrapped in the one raised.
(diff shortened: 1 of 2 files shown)


gnssanalysis/gn_download.py
```diff
@@ -903,59 +931,172 @@ def download_product_from_cddis(
...
+    if_file_present: str = "prompt_user",
+) -> Union[_Path, None]:
+    """
+    Compatibility wrapper around new functions
+    DEPRECATED since approximately version 0.0.58
+    """
+    variants = get_iau2000_file_variants_for_dates(start_epoch=start_epoch)
+    if len(variants) != 1:
+        raise NotImplementedError(
+            "Legacy wrapper for IAU2000 file download failed. Exactly one file variant should be returned based on "
```
ronaldmaj, line 1095: I seem to hit this error whenever I put through a start date older than 90 days:
```py
In [23]: download_iau2000_file(download_dir=dwn_dir, start_epoch=st2)
NotImplementedError                       Traceback (most recent call last)
Cell In[23], line 1
----> 1 download_iau2000_file(download_dir=dwn_dir, start_epoch=st2)

File /data/gnssanalysis/gnssanalysis/gn_download.py:1094, in download_iau2000_file(download_dir, start_epoch, if_file_present)
   1092 variants = get_iau2000_file_variants_for_dates(start_epoch=start_epoch)
   1093 if len(variants) != 1:
-> 1094     raise NotImplementedError(
   1095         "Legacy wrapper for IAU2000 file download failed. Exactly one file variant should be returned based on "
   1096         f"a single date ({start_epoch})."
   1097     )
   1098 variant = variants.pop()
   1099 download_iau2000_variant(download_dir=download_dir, iau2000_file_variant=variant, if_file_present=if_file_present)

NotImplementedError: Legacy wrapper for IAU2000 file download failed. Exactly one file variant should be returned based on a single date (2024-10-23 06:00:00).
```

When I run the `get_iau2000_file_variants_for_dates()` function with just the `start_epoch` above, I get back two variants (rather than one)

gnssanalysis/gn_download.py
```diff
@@ -903,59 +931,172 @@ def download_product_from_cddis(
...
+            )
+            if download_filepath_or_none is None:  # File already existed and was not re-downloaded.
+                return None
+            # Download succeeded. Rename to standard filename, and return the path
+            return download_filepath_or_none.rename(download_dir / download_filename)
+        except Exception as ex:
+            logging.error("Failed to download IAU2000 file from CDDIS.", ex)
+
+
+def get_iau2000_file_variants_for_dates(
```
ronaldmaj, line 1008: When I run this with a single date (a `start_epoch`) a little over 90 days ago, I get two variants back, rather than one

ronaldmaj (changes requested):
I've discovered a couple errors coming up that I think shouldn't be, based on my inputs?
Something to look at next week perhaps

gnssanalysis/gn_download.py
treefern: Good catch, thanks. I should've put in a specific test for this wrapper.

treefern: Logic and tests updated. Still no test for this specific function (as it also handles the download), but there is a test for running the variant picker in the legacy mode this wrapper now uses.

gnssanalysis/gn_download.py
treefern: This should now be addressed for the legacy wrapper `download_iau2000_file()`.

For the new function, this is expected behaviour. For a range that doesn't specify an end, we consider the possibility that it runs up to the current date. Therefore, we include the 'daily' file to be safe.

gnssanalysis/gn_download.py
ronaldmaj: I've tested this on dates less than and greater than a week / 8 days. Seems to work well, downloading the daily file within 8 days and the long term data file for dates beyond that (as we previously had it for this function)

ronaldmaj (approved):
This is looking good to go in now. My tests of the legacy function work well now, so no more concerns from me :)

PR #75, opened 2025-03-04
NPI-3781 Fix crash on SP3 header update after Helmert transform

treefern:
A mistake in header update code added in December, causes a crash when trying to update the SP3 header after running a Helmert transform.


seballgeyer (approved):
makes sense. 

Thanks

PR #76, opened 2025-03-06
NPI-3793 Fix misleading operator use on sets

treefern:
There are some instances where a bitwise AND operator appears to be used on structures for which it does not make sense.

Plus add clarifying comment for some Numpy custom operator usage.

This PR was split out from work done to restructure clk comparison and sisre calculation. TODO: a small fix is needed.


treefern:
Closed this PR as the relevant code is downstream and will be merged in via https://github.com/GeoscienceAustralia/gnssanalysis/pull/77

PR #79, opened 2025-04-24
NPI-3918 Missing value check robustness improvements, and various type hint fixes

treefern:
Various robustness improvements to missing value checks, type hints and mismatches. Most significant changes in sinex.py


ronaldmaj (approved):
This looks good, making everything explicit will ensure we don't get stung by weird edge cases. I've tested this by using the `download_atx()` function to see if the `attempt_url_download()` is properly called and used.
There could be more extensive testing done to ensure there are no major changes, but for the single test I did, it worked as expected

PR #82, opened 2025-05-06
NPI-3945 Bugfix and further cleanup of GPSDate

treefern:


ronaldmaj (approved):
Awesome, thanks for getting this in so quickly. Also, I think the new name is great `internal_dt64` 👍

PR #83, opened 2025-05-07
NPI-3973 Better error logging for SType checks

treefern:
We are seeing odd issues which are crashing the SType check. This PR adds a bit more logging around that.

This PR also restores a bizarre and previously removed code path in `_get_snx_vector()`, and logs an error if it is used.


ronaldmaj (approved):
This will help debug the issues we are seeing on ginan-ops, so happy for this to go in 👍

PR #84, opened 2025-05-08
NPI-3973 Fix bug in specification of stypes

treefern:
Set containing a single string, mistakenly defined as a set of characters within that string. This is likely causing valid stypes to be rejected.


ronaldmaj (approved):
This minor change to the code that fixes a bug coming through in our ginan-ops code. Happy for this to go in 👍

PR #85, opened 2025-05-15
NPI-3999 Fix clean_sp3_orb

treefern:
Unit test to follow soon


ronaldmaj (approved):
Looks good 👍 
I've tested this locally with an example file. Current release does not work, whereas this fix passes through fine.

PR #89, opened 2025-05-16
Npi 3470 misc clarity and hint improvements

treefern:
Some misc cleanup work from August last year


ronaldmaj (changes requested):
Couple minor changes that will need to be made to allow this to go in

ronaldmaj (dismissed):
Cool, looking great now. Thanks for pushing through those changes

PR #91, opened 2025-06-30
NPI-4066 Reject SP3 files with overlong data lines

treefern:
SP3 data lines are fixed width and strictly follow a columnar layout. Misaligned data can lead to serious parsing issues, including grossly incorrect position values.

This PR adds a simple check to `read_sp3()`, rejecting any SP3 data which has line(s) exceeding 80 chars.

Further work is in train to more robustly check alignment by testing that all unused columns contain a space.


ronaldmaj (approved):
This is a nice PR that quickly tests for lines are too long, which is usually indicative of Ginan processing gone awry. It'll enable our rapid and ultra-rapid jobs to fail quickly when there are issues.

I've tested this with a file that was problematic - it detected all lines that exceeded the line length successfully.

Happy for this to go in 👍

PR #93, opened 2025-07-11
NPI-4067 Improvements to SP3 validation and unit tests

treefern:
This PR improves content validity / consistency tests in SP3 reading, tidies up unit tests, and addresses some Pandas deprecation and performance warnings.

- Checks for SP3 column alignment (looks for spaces in columns which the (SP3d) spec declares as unused)
- Revised `strict_mode` logic and structure of options, including optional overrides for specific use cases (e.g. comments, duplicate epochs, content vs header discrepancy).
- Various unit test clean ups, including:
  - Using bytes input rather than mock files for input in many SP3 tests (simpler).
  - Fixes for checking values of raised exceptions (some checks were previously unreachable and have been moved outside the `assertRaises()` `ContextManager`).
  - Quieter output, by turning off `strict_mode` for things a test is not trying to actively check.
- Pandas deprecation fixes including `.view()` -> `.astype()`
- Performance warning fixes, by sorting DataFrame and by specifying a level when dropping.

### DIFF
--- a/tests/test_clk.py
+++ b/tests/test_clk.py
@@ -34,8 +34,10 @@ class TestClk(TestCase):
         self.assertEqual(clk_df_igs["EST"].iloc[-1], -0.0006105557076344, msg="Check last datapoint is correct")
         self.assertEqual(clk_df_gfz["EST"].iloc[-1], -0.000610553573006, msg="Check last datapoint is correct")
 
-    # TODO replace with test for revised diff_clk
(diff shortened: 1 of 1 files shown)


treefern:
Previous test issue fixed.
A few other improvements added including:
- more backwards compatible defaults for filename property extraction
- more reliable discrepancy check
- quieter error logging for overlong SP3 content lines
- better docstrings

PR #95, opened 2025-09-25
Switch CDDIS downloads from FTP-SSL to HTTPS with NASA Earthdata authentication

aclel:
This PR migrates CDDIS file downloads from FTP-SSL to HTTPS using NASA Earthdata authentication, providing better reliability, security, and performance for downloading GNSS products.

## Migration Guide

  Users should:
  1. Set up NASA Earthdata credentials in ~/.netrc with entry for urs.earthdata.nasa.gov
  2. Update code to use url_folder parameter instead of ftp_folder

###  Modified Functions

- New authentication system: Added get_earthdata_credentials() function to handle NASA Earthdata login via .netrc file or direct credentials
- download_file_from_cddis(): Rewrite to use HTTPS with requests.Session and NASA Earthdata auth
- download_multiple_files_from_cddis(): Updated to use new HTTPS approach with thread pool
- download_product_from_cddis(): Uses new HTTPS approach

 ### API Changes

  - Added url_folder parameter (preferred) while keeping ftp_folder as deprecated
  - Both functions now accept username/password parameters for NASA Earthdata authentication
  - Maintained backward compatibility with existing parameter names

### Dependencies

  - Added requests library for HTTPS downloads

## Breaking Changes
 See Migration Guide above for authentication requirements. All changes are backward compatible. The deprecated ftp_folder parameter still works but issues a deprecation warning.

### DIFF
--- a/gnssanalysis/gn_download.py
+++ b/gnssanalysis/gn_download.py
@@ -28,7 +28,7 @@ import hatanaka as _hatanaka
 import ftplib as _ftplib
 from ftplib import FTP_TLS as _FTP_TLS
 from pathlib import Path as _Path
-from typing import Any, Generator, Literal, Optional, Union
+from typing import Any, Generator, Literal, Optional, Union, Tuple, List
 from urllib import request as _request
 from urllib.error import HTTPError as _HTTPError, URLError as _URLError
 import requests as _requests


ronaldmaj (approved):
This is a great contribution, and something we've been meaning to do for a while. I've tested it locally and seems to work well.
It's been able to download SP3, CLK, SNX, BIA files from CDDIS, both IGS and COD. I've also tried both ULT and RAP (some issues with CLK and BIA as they don't have ULT versions, but that's an issue that currently exists in the code and not caused by the incoming code).
Happy for this to go in 👍

PR #100, opened 2025-10-21
NPI-4271 Support 01W or 07D to express nominal span of one week

treefern:


ronaldmaj (approved):
Looks good, tested this locally and works as expected. The 01W vs 07D option here is great so that users can choose between the two (defaulting to follow IGS)

PR #101, opened 2025-10-21
NPI-4237 Cleanup for Python3.9 retirement

treefern:


ronaldmaj (approved):
Python 3.9 is now deprecated so we definitely need to remove it from our tests and bring in new conventions on type hints - happy for this to go in 👍

PR #106, opened 2026-01-22
NPI-4377 Small revisions to NANU processing utility functions

treefern:
Ahead of more significant NANU processing revisions, this PR adds some general robustness and clarity improvements to existing utility functions that support NANU processing.

This should have negligible impact on how things currently work.

⚠️ **Note: due to breaking changes in Pandas 3, #109 will need to be merged first.**

### DIFF
--- a/gnssanalysis/gn_io/nanu.py
+++ b/gnssanalysis/gn_io/nanu.py
@@ -1,8 +1,8 @@
 import glob
 import logging as _logging
 import os as _os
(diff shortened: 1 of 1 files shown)


treefern:
The test failures here look unrelated to these changes.
They appear to be breaking changes in `pandas 3.0.0` which we had not fully addressed ahead of time. I'll make a separate PR to resolve these. Please see #109

PR #110, opened 2026-02-05
NPI-4453 Framework for DataFrame hashing & test baselining

treefern:
Introduces a framework for baselining lists of `DataFrames` (and in future, other object types) produced by unit tests, then checking for regression against these in subsequent runs.

### Baselining mode
The exact `DataFrame` outputs of a unit test can be 'baselined' within unit tests:
```
df_1 = load_some_data()
df_2 = transform_something()
self.assertEqual(df_1, some_value)
...
df_list = [df_1, df_2]
UnitTestBaseliner.mode = "baseline"
UnitTestBaseliner.create_baseline(df_list)
```
The baseline is comprised of two files (a sha256 hash, and a pickled `list[object]`) which can be committed along with relevant changes.
To prevent accidental baselines being created, `UnitTestBaseliner.mode = "baseline"` must be set, which turns off `verify` mode and raises warnings to reduce the risk of this state being committed.

### Verification mode
Subsequent unit test runs can call `UnitTestBaseliner.verify()`, passing a `list[object]`, of the outputs they have produced.
```
df_1 = load_some_data()
df_2 = transform_something()
self.assertEqual(df_1, some_value)
...
df_list = [df_1, df_2]
UnitTestBaseliner.verify(df_list)
```
This compares the current unit test output against the baseline on file (using just hash for detection).
> This will allow detection of regressions too subtle to be found by our existing unit tests.

#### Troubleshooting regressions with DataFrame diffs
When a unit test invokes `verify()` and it fails (hash not valid), the `verify()` function can load the baselined `list[object]` from the pickle file, and print diffs between these, and the current output of the unit test.
- Currently this is only supported for `Dataframe` and will fail if other data types are included.
> ⚠️ NOTE: Due to the security implications of deserialization, it must be explicitly enabled when needed, with `UnitTestBaseliner.enable_unpickling = True`

### Baseline file storage
Baseline files are stored at:
`gnssanalysis/tests/unittest_baselines/<class_name>/<unittest_function_name>.{pickledlist,pickledlistsha256}`

**Note:** When `create_baseline()` or `verify()` is invoked, the names of the calling class and function are determined automatically using frame inspection.

This means simply invoking them from within a unit test, will cause a corresponding directory and baseline files to be written or read at the path noted above.


tests/test_utils.py
```diff
@@ -64,3 +67,129 @@ def test_configure_logging(self):
...
+
+
+class TestUnitTestBaseliner(unittest.TestCase):
+
+    def test_verify_refusal_in_wrong_mode(self):
+        mode_backup = UnitTestBaseliner.mode
+        try:
+            df = DataFrame(["a", "b", "c"])
+
+            # Baseline (do not commit uncommented!) Note: every function needs its own baseline, becuase the
```
ronaldmaj, line 79: typo: `becuase`

tests/test_utils.py
```diff
@@ -64,3 +67,129 @@ def test_configure_logging(self):
...
+        # These functions determine what files to write/read baselines from, based on the identity of the (test)
+        # function that called them. Therefore, calling twice from the same function would cause the *same baseline
+        # files* to be read/written for a different part of the unit test.
+        # That would have the effect of:
+        # - in write mode: overwriting the baseline file for a previous part of the test function.
+        # - in read mode: repeating verification of the same file against a different DF / object list (which would
+        #   likely fail).
+
+        # We're only testing it with the verify function below, but both verify and baseline functions use the same
+        # caller check logic, and store the caller record statically in a class variable. ?
```
ronaldmaj, line 124: Is the question mark here for a reason?

tests/test_utils.py
```diff
@@ -64,3 +67,129 @@ def test_configure_logging(self):
...
+        # UnitTestBaseliner.mode = "baseline"
+        # UnitTestBaseliner.record_baseline(dfs_to_hash)
+
+        # Will return True if verification succeeded. False if baseline missing or mode != verify
+        self.assertTrue(
+            UnitTestBaseliner.verify(objects_to_hash),
+            "DF / object list verification should succeed here (unless baseline files are missing, or baselining has been turned on)",
+        )
+
+        # The local variable df still points to the same DF, so now the list contains [a,b,b]. This should be an error.
```
ronaldmaj, line 166: Not sure where the dataframe `[a,b,b]` came from? You mean `["b", "c", "d"]` which is what the `df` var now points at?

tests/test_utils.py
```diff
@@ -64,3 +67,129 @@ def test_configure_logging(self):
...
+        # - in write mode: overwriting the baseline file for a previous part of the test function.
+        # - in read mode: repeating verification of the same file against a different DF / object list (which would
+        #   likely fail).
+
+        # We're only testing it with the verify function below, but both verify and baseline functions use the same
+        # caller check logic, and store the caller record statically in a class variable. ?
+
+        df = DataFrame(["a", "b", "c"])
+
+        # Baseline (every function needs its own baseline, becuase the function name determines the filename,
```
ronaldmaj, line 128: typo: `becuase`

tests/test_utils.py
```diff
@@ -64,3 +67,129 @@ def test_configure_logging(self):
...
+        # UnitTestBaseliner.record_baseline(dfs_to_hash)
+
+        # Will return True if verification succeeded. False if baseline missing or mode != verify
+        self.assertTrue(
+            UnitTestBaseliner.verify(objects_to_hash),
+            "DF / object list verification should succeed here (unless baseline files are missing, or baselining has been turned on)",
+        )
+
+        # The local variable df still points to the same DF, so now the list contains [a,b,b]. This should be an error.
+        objects_to_hash.extend([df])
```
ronaldmaj, line 167: You are trying to add the same dataframe here to `objects_to_hash` and that is what is going to cause the error when verifying right? Because you have a duplicate?

tests/test_utils.py
```diff
@@ -64,3 +67,129 @@ def test_configure_logging(self):
...
+        self.assertTrue(
+            UnitTestBaseliner.verify([df]),
+            "DF / object list verification should succeed on *first* call from a function.",
+        )
+        with self.assertRaises(ValueError):
+            UnitTestBaseliner.verify([df])
+            self.fail("DF / object list verification should fail on *second*/repeated calls from a function.")
+
+    def test_duplicate_object_rejection(self):
+
```
ronaldmaj, line 142: No description of the test here, whereas the other previous has an intro to what is being tested

tests/test_utils.py
```diff
@@ -64,3 +67,129 @@ def test_configure_logging(self):
 
         # Verify
         self.assertEqual(logger_not_output, None)
+
+
+class TestUnitTestBaseliner(unittest.TestCase):
+
+    def test_verify_refusal_in_wrong_mode(self):
```
ronaldmaj, line 74: Short intro to what is being tested would be good here (kinda implied in the name, but could be good to add in)

ronaldmaj (approved):
Did a quick read through the code, high-level understand that this introduces a framework for having data which we can regression test against - if answers change, we need to investigate.

I think this is a nice framework, as it allows us to update the hashes (and pickled versions of the data) in case the new values are actually better - having to do manual overwrite is a feature. 

Ran the tests locally as well, and in it's current form works fine.
I peppered in some comments but it's nothing major and I think it is good to go in it's current form 👍

tests/test_utils.py
treefern: Yep. This checks for duplicate references to the same objects at the top level (as a safety check). It's not recursive, but the top-level check is arguably the most important.

tests/test_utils.py
treefern: That was intended to be shorthand for the different DataFrame objects, rather than their content.
This check doesn't care about the data the object stores, just the object's memory address.

tests/test_utils.py
treefern: I wasn't sure on the specific terminology. Checked and updated with a clarification.

treefern:
Minor changes suggested above, are now on: #111

PR #114, opened 2026-03-06
NPI-4495 Earthdata creds via env var

treefern:
A NASA Earthdata account is required for some file downloads.
This PR adds an additional option to the credential fetching function for this, allowing creds to be provided via env vars.


ronaldmaj (approved):
Looks good - allows this to search for environment variables that assign the EarthData credentials. 
Tested this locally and works as expected picking up the environment variables.

This also means that environment variables override any `.netrc` present in home directory

PR #116, opened 2026-06-29
NPI-4711 Simple extraction of IGS long filename properties

treefern:


ronaldmaj (approved):
This looks good to me - introducing a simple function to allow verbatim extraction of the various parts of the filename

